PR #11, opened 2023-11-23
Typography/Font

vishkaischibsted:
Introduce new type named `Typography` which contains semantic fonts associated with font weight, that has been adopted from `Warp` design system logics.


SampleApp/Warp/WarpApp.swift
```diff
@@ -0,0 +1,15 @@
+import SwiftUI
+import Warp_ios
+
+@main
+struct WarpApp: App {
+    init() {
+        JBMono.registerFonts()
```
mohsen-schibsted, line 7: @vishkaischibsted didn't you remove this?

vishkaischibsted: It's in a removed app target.

PR #15, opened 2024-01-24
Add badge component

mohsen-schibsted:


vishkaischibsted (approved):

Sources/Badge/Badge.swift
```diff
@@ -0,0 +1,117 @@
+import SwiftUI
+
+extension Warp {
+    /// Use badges to highlight status of an ad (Active/Inactive/Sold/Removed) or to mark paid placements (Sponsored/Ad/House of the week).
+    public struct Badge: View {
```
vishkaischibsted, line 5: If you see it appropriate, conform the whole view to `Hashable` in order to prevent unnecessary updates.
Feel free to ignore it since it's only a sugguestion.

PR #18, opened 2024-01-25
Alert

vishkaischibsted:
Add new view named `Alert`, based on [Figma](https://www.figma.com/file/8P1JQsd82b93gQ6K3igO2p/Warp---Components?node-id=171%3A20423&mode=dev) designs,
Alert has 4 types that can be found under Warp name space, an enum named `AlertStyle`.

# Simple use case:

simulator_screenshot_1CC60B1E-28CD-44D2-A899-F0E827858208

# Full use case:

simulator_screenshot_6A3AD7DD-888B-433C-93F6-B9197A3A2B78


SampleApp/AlertView.swift
```diff
@@ -0,0 +1,53 @@
...
+                    secondaryButtonProvider: (title: "Button", action: {})
+                )
+                .padding()
+            }
+        }
+    }
+}
+
+extension Warp.AlertStyle: CaseIterable {
+    public static var allCases: [Warp.AlertStyle] = [
```
mohsen-schibsted, line 43: why not use `CaseIterable` protocol directly?

Sources/Alert/Alert.swift
```diff
@@ -0,0 +1,366 @@
...
+            .frame(height: 1)
+            .offset(y: 8)
+    }
+}
+
+extension Warp.AlertStyle {
+    fileprivate func getBackgroundColor(from colorProvider: ColorProvider) -> Color {
+        switch self {
+            case .info:
+                return FinnColors.aqua50
```
mohsen-schibsted, line 274: We should NOT use Brand colors directly inside any components, that way it will not change color if theme is different. In this case use colorProvider.alertInfoBackground

Sources/Alert/AlertStyle.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+
+extension Warp {
+    /// <#Description#>
+    public enum AlertStyle: Equatable {
+        /// <#Description#>
```
mohsen-schibsted, line 6: I would copy and paste the descriptions from Figma to here.
Screenshot 2024-01-26 at 12 29 49

SampleApp/AlertView.swift
vishkaischibsted: I am using it directly here :!
Can I ask how you mean I should use it?

Sources/Alert/Alert.swift
vishkaischibsted: Thanks for your comment, how should I use them?
Since they do not exists in the colors right now, can I add them in color provider and then add them here?
Or I have to wait until it's added?

Sources/Alert/AlertStyle.swift
vishkaischibsted: Fantastic, thanks!

vishkaischibsted: Can I ask where it is in Figma?
I can't find it in [Alert](https://www.figma.com/file/8P1JQsd82b93gQ6K3igO2p/Warp---Components?type=design&node-id=171-20423&mode=design&t=PvP14KlqRUqXVjRw-0) section!

mohsen-schibsted (approved):

PR #19, opened 2024-01-31
feat: Adds suggestion for a Warp Toast component

pbodsk:
# What?
Here you will find a suggestion for a *Toast* component matching what we have in the Warp design system.

## Examples
Let's start with a couple  of images so you can see what we're working with here.

Screenshot 2024-01-31 at 09 56 19

Screenshot 2024-01-31 at 09 56 31

## Usage
You need to consider the following options when using the Toast.
### Title
You need to provide the toast with a `title`. For now we only support a simple `String` value but we could add support for `View` content if so needed.

### Style
You have the option to style the toast as:

- an Error toast
- a Success toast
- a Warning toast

by using the appropriate `ToastStyle` enum value

### Location
Furthermore you can control the location of the toast. You can place it:

- at the Top of the screen
- at the Bottom of the screen

by using the appropriate `ToastEdge` enum value.

I have added a `ToastView` to the example app that illustrates how to use the toast view and how you can use a `viewModifier` to control showing/hiding of the toast.

## Design
The Toast feature is located in "Sources/Toast" and is "wrapped" in the `Warp` namespace.

It contains of:

- A `Toast` SwiftUI view where we render the actual component.
- A `ToastStyle` enum to control the colors and icons of the various styles
- A `ToastEdge` enum to control where to show the toast
- A `ToastViewModifier` that allows the user to easily attach a Toast component to their View

I've tried to make the naming and file structure as close to the existing `Alert` feature as I could, just so there is consistency (I hope 😅)

## Missing Features
- The Toast will not automatically dismiss after a pre-defined time interval, I'm not sure that is something we need to support?
- The Toast only supports appearing at the top or bottom of the view

## What I Would Like Your Feedback On
- What do you think about the format, is it what you would like from a Warp component?
- How do we handle icons? Right now I've added new icons but is that how we want to solve this?
- What do you think about the ViewModifier, should we go with a different name?

## Preview Bug
I would _love_ to be able to run my Toast view in preview as I build it, but I get an error when doing so...the same seems to be the case for the Alert view, it complains that `Compiling failed: cannot find 'Typography' in scope`. Is that something we can find a workaround for?

Have fun 😄


Sources/Toast/ToastStyle.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+extension Warp {
+    public enum ToastStyle: Equatable, CaseIterable {
+        /// Indicate a severe problem or error.
```
mohsen-schibsted, line 5: Please use these for documentations, you can find them [here](https://www.figma.com/file/nkiRpuVu6XRfvY96BA80H8/Components-overview?type=design&node-id=377-23910&mode=design&t=Xt8LzafYSVzUNSGi-0)
This way we will sync our documentation across platforms, unless you have copied them from somewhere that makes more sense.
Screenshot 2024-01-31 at 13 03 51

Sources/Toast/Toast.swift
```diff
@@ -0,0 +1,142 @@
...
+            colorProvider.toastPositiveText
+        case .warning:
+            colorProvider.toastWarningText
+        }
+    }
+
+    fileprivate var icon: Image {
+        switch self {
+        case .error:
+            Image("icon_toast-error", bundle: .module)
```
mohsen-schibsted, line 126: These images need tint color from colorProvider too. As you can see in the image attached they should be different in Finn and Tori. Probably it is toast***Icon
Screenshot 2024-01-31 at 13 06 10

mohsen-schibsted: The background colors do not match either, specially on Tori but it seems you have used the right colors so probably something wrong on Figma. I will investigate that.
Screenshot 2024-01-31 at 13 06 20

mohsen-schibsted: And for the icons is you have the same problem as the one in the Alert (the icon does not have transparent exclamation mark) then you can not use tint probably. Just do it like Alert (use systemIcons) until we get correct icons.

And also, the icons seems duplicate with the ones we have for Alert. So maybe we can reuse them, you can find the icons [here](https://www.figma.com/file/1qkEGDQWaftkOL7C360qVD/Fabric---Icon-library?type=design&node-id=3482-2577&mode=design&t=OAQYXYQQSne58IWo-0) and they are called Alert, so should be okey to reuse later on when we have the correct icons.

mohsen-schibsted: Yes the colors on Figma is wrong and our designers are fixing that, so we can ignore that. But the icon need taking care of. And the border, it seems ours is a little bit narrow, I can see you have used 2 just like figma, but could you please make it bigger and see if it looks more like figma?

Sources/Toast/ToastStyle.swift
pbodsk: Ah nice! Then I don't have to come up with descriptions myself when someone who actually knows what they are doing has already done the work 😅

Sources/Toast/Toast.swift
pbodsk: > But the icon need taking care of.

Sure thing ✅ 

I've changed so we now use the system icons and delete my Toast variants. Then once we have the right Alert icons I can re-use those (and then maybe we need to come up with different names to avoid confusion about why we're using "Alert" icons in the "Toast" for instance).

pbodsk: > And the border, it seems ours is a little bit narrow, I can see you have used 2 just like figma, but could you please make it bigger and see if it looks more like figma?

I've attached two screenshots here:

**one where I've used a `lineWidth` of 3px**

lineWidth 3px

**and one where I've used a `lineWidth` of 4px**

lineWidth 4px

I have a hard time deciding which I prefer but I _think_ I like the 4px version best. Ask me again tomorrow and I may say 3px 😅

mohsen-schibsted: I think 4 looks more like the one on Figma, so maybe the way the borderWidth is setup on Figma is different. I will bring it up with my team.

And regarding using Alert icon name in Toast component we will use whatever Figma tells us to use. We will always have Figma as source of truth.
Screenshot 2024-02-01 at 09 46 21

Sources/Toast/Toast.swift
```diff
@@ -0,0 +1,163 @@
...
+            contentView
+            .overlay(
+                RoundedRectangle(cornerRadius: toastCornerRadius)
+                    .stroke(style.subtleBorderColor(from: colorProvider), lineWidth: 3)
+            )
+            .frame(maxWidth: .infinity)
+            .background(style.backgroundColor(from: colorProvider))
+            .cornerRadius(toastCornerRadius)
+            .transition(AnyTransition.move(edge: toastEdge.asEdge).combined(with: .opacity))
+            .onTapGesture {
```
mohsen-schibsted, line 68: TapGesture is added to the whole background, is that intentional? We don't want that only on the close button? There is no mention of that on Figma. I have asked our designers and they said there shouldn't be a close button there at all and it should auto dismiss because of a11y reasons. They will look into it, so we do not need to do anything regarding this right now. We will fix that later on.

mohsen-schibsted (approved):

mohsen-schibsted:
LGTM, thank you for helping out 🥳

Sources/Toast/Toast.swift
mohsen-schibsted: our designers agreed on 4 too

pbodsk: > our designers agreed on 4 too

Right, 4 it is, I'll update ✅

Sources/Toast/Toast.swift
pbodsk: > TapGesture is added to the whole background, is that intentional?

It actually was intentional 😄 

I like that you don't have to try to hit a tiny close button to dismiss but can just tap wherever on the view to dismiss. In the ideal world I'd probably also added a swipe gesture recognizer so you could swipe up to dismiss but that may be going overboard 😅 

> They will look into it, so we do not need to do anything regarding this right now. We will fix that later on.

OK we'll come back here and add a timer it seems 😄

pbodsk:
> LGTM, thank you for helping out

Of course. Making UI  components is fun, especially in SwiftUI 😄 

I have a `Callout` like component as well that I'll see if I can get polished and added as well....to be continued!

PR #21, opened 2024-02-05
Feature/callout

pbodsk:
I had the need for a `Callout` component as well so I went ahead and made something that I hope can be a starting point for a Warp Callout component.

## Examples

First a couple of examples

Simulator Screenshot - iPhone 15 - 2024-02-05 at 13 42 49 Medium
Simulator Screenshot - iPhone 15 - 2024-02-05 at 13 42 55 Medium

## Usage
You need to provide the `Callout` with:

- Title: A `String` value to show in the callout
- arrowHeight (optional): Height of the arrow to draw, default value is 8 px
- arrowWidth (optional): Width of the arrow to draw, default value is 18 px
- arrowEdge (optional): Edge where to draw the arrow, default value is `.top`
 - cornerRadius (optional): Radius of corners, default value is 8

I have added a `CalloutView` to the example app that illustrates how to use the toast view

## Design
The Callout feature is located in "Sources/Callout" and is "wrapped" in the Warp namespace.

It consists of:

- A `Callout` SwiftUI view where we render the actual content (take the title, put it in a `Text`, style it and add a `CalloutShape` as the `.background` to the  `Text`)
- A `CalloutShape` `Shape` where we do the calculations and drawing needed to actually draw the callout shape. This seems a bit hairy at first with a lot of draw commands. I've tried documenting it so it is - I hope - easier to dive into in six months 😅 

## Missing Features
I have not yet implemented the "popover" variant that adds a button and a shape. I can do so if you'd like me to, I just didn't have the need for it myself when I made this.

## What I Would Like Your Feedback On
- As in #19, the format
- Are we good with the `CalloutShape` being what it is now? (quite a lot of manual drawing)
- Do I need to lock arrowWidth, height etc. so they can't be changed from the outside but is just defined in the component?
- Component design...is it "user-friendly" and understandable as a "component iOS developer consumer"

As last time, have fun...and thank you for your feedback 😄


Sources/Callout/Callout.swift
```diff
@@ -0,0 +1,146 @@
...
+            self.arrowHeight = arrowHeight
+            self.arrowWidth = arrowWidth
+            self.arrowEdge = arrowEdge
+            self.cornerRadius = cornerRadius
+            self.colorProvider = colorProvider
+        }
+
+        public var body: some View {
+            Text(title)
+                .font(from: Typography.body)
```
mohsen-schibsted, line 65: font is different with the size too, seems one is using body the other caption

mohsen-schibsted: and seems size small is only one liner but size default support multi line text

Sources/Callout/Callout.swift
```diff
@@ -0,0 +1,146 @@
...
+                .foregroundStyle(colorProvider.calloutText)
+                .padding(.horizontal, 12)
+                .padding(.vertical, 9)
```
mohsen-schibsted, line 68: paddings depend on size too

Sources/Callout/Callout.swift
```diff
@@ -0,0 +1,146 @@
...
+    To use the iOS `Callout` View you need to provide it with
+     - a `title`
+     - an `arrowHeight`, if you don't provide this, the default value is 8 px
+     - an `arrowWidth`, if you don't provide this, the default value is 18px
+     - an `arrowEdge` where you want the Callout to appear from
+     - a `cornerRadius`, if you don't provide this, the default value is 8px
+     */
+    public struct Callout: View {
+        /// Title to present in the callout view
+        private var title: String
```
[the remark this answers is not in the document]
pbodsk: Ha! I missed that there were two different size 😅 

I'll hide the arrowWidth/Height and introduce enums for `size` and `type`

Sources/Callout/Callout.swift
pbodsk: I see an `enum` with computed properties in my near future. God I love `enum`s in Swift 🤓

Sources/Callout/Callout.swift
pbodsk: Fixed

pbodsk:
@mohsen-schibsted There we go, we can now support various `size`s and `type`s of `Callout`s. I've updated the example.

## A note on `CalloutType`

I can't figure out if this too clever

```
public enum CalloutType {
    public enum PopoverType {
        case dismissable(onTapped: () -> Void)
        case nondismissable
    }

    case inline
    case popover(type: PopoverType)
}
```

I mean, it gets the job done and it is quite elegant I _think_, but I don't know if it is too difficult to figure out how to use it.

Here's how it looks on the callsite

```
Warp.Callout(
    type: .popover(
        type: .dismissable(onTapped: {
            isVisibleValue.wrappedValue = false
        })
    ),
    title: "Some title",
    arrowEdge: .top
)
```

Or 

```
Warp.Callout(
    type: .popover(
        type: .nondismissable
    ),
    title: "Some title",
    arrowEdge: .top
)
```

or if you just want the default inline variant you don't need to pass in any type, unless you want, and in that case it looks like so:

```
Warp.Callout(
    type: .inline,
    title: "Some title",
    arrowEdge: .top
)
```

What do you think?

mohsen-schibsted:
> @mohsen-schibsted There we go, we can now support various `size`s and `type`s of `Callout`s. I've updated the example.
> 
> ## A note on `CalloutType`
> I can't figure out if this too clever
> 
> ```
> public enum CalloutType {
>     public enum PopoverType {
>         case dismissable(onTapped: () -> Void)
>         case nondismissable
>     }
> 
>     case inline
>     case popover(type: PopoverType)
> }
> ```
> 
> I mean, it gets the job done and it is quite elegant I _think_, but I don't know if it is too difficult to figure out how to use it.
> 
> Here's how it looks on the callsite
> 
> ```
> Warp.Callout(
>     type: .popover(
>         type: .dismissable(onTapped: {
>             isVisibleValue.wrappedValue = false
>         })
>     ),
>     title: "Some title",
>     arrowEdge: .top
> )
> ```
> 
> Or
> 
> ```
> Warp.Callout(
>     type: .popover(
>         type: .nondismissable
>     ),
>     title: "Some title",
>     arrowEdge: .top
> )
> ```
> 
> or if you just want the default inline variant you don't need to pass in any type, unless you want, and in that case it looks like so:
> 
> ```
> Warp.Callout(
>     type: .inline,
>     title: "Some title",
>     arrowEdge: .top
> )
> ```
> 
> What do you think?

I think it is totally clever and understandable. The thing is I think `Callout` type `.popover` is always `dismissable`. There is always gonna be an `X` there to close it. So there won't be a `nondismissable` `.popover`. Shadows always come with a close button according to [this](https://www.figma.com/file/8P1JQsd82b93gQ6K3igO2p/Warp---Components?type=design&node-id=381-42328&mode=design&t=CTAoXRY84fFkOl86-0).

Sources/Callout/CalloutSize.swift
```diff
@@ -0,0 +1,35 @@
...
+        public var cornerRadius: Double {
+            switch self {
+            case .small:
+                6
+            case .default:
+                8
+            }
+        }
+
+        public var verticalPadding: Double {
```
mohsen-schibsted, line 26: I think you are not using this anywhere. That's why the height looks a little bit smaller than design.
Simulator Screenshot - iPhone 15 - 2024-02-07 at 18 27 06
Screenshot 2024-02-07 at 18 27 34

mohsen-schibsted: At least the ones with vertical arrows, the leading and trailing seems correct.

pbodsk:
> Shadows always come with a close button according to [this](https://www.figma.com/file/8P1JQsd82b93gQ6K3igO2p/Warp---Components?type=design&node-id=381-42328&mode=design&t=CTAoXRY84fFkOl86-0).

🤔 you're right. I just saw some popovers without close buttons somewhere but that must've been a mistake...and it make things easier.

I'll align so that popovers always has an `onDismiss` callback 👍

pbodsk:
@mohsen-schibsted 

Can you help me figure out if the horizontal padding should be 12 or 16 px?

There seems to be two opinions in the Figma design

In the Components Overview it says 16px

Screenshot 2024-02-08 at 09 45 35

And in the actual Warp Components page it says 12px

Screenshot 2024-02-08 at 09 46 01


🤔

mohsen-schibsted:
> @mohsen-schibsted
> 
> Can you help me figure out if the horizontal padding should be 12 or 16 px?
> 
> There seems to be two opinions in the Figma design
> 
> In the Components Overview it says 16px
> 
> Screenshot 2024-02-08 at 09 45 35
> 
> And in the actual Warp Components page it says 12px
> 
> Screenshot 2024-02-08 at 09 46 01
> 
> 🤔

I can see in the place that it is 12, there is a gap which is 8 too 😅 so probably 16 is the correct one. I will ask our designers too.
Screenshot 2024-02-08 at 10 27 38
Screenshot 2024-02-08 at 10 28 00

mohsen-schibsted (approved):

mohsen-schibsted:
LGTM, the height when the arrow is on the top or bottom seems a bit low compared to design, but the overall component works as expected. So feel free to fix that then merge or merge and then we will fix that later on design review.

pbodsk:
Got the vertical padding fixed...as discussed on Slack 😄

PR #22, opened 2024-02-06
fix: Adding padding to the toast no longer updates main layout

pbodsk:
## Problem
I discovered that by setting the `.padding(.horizontal, 16)` in our ViewBuilder we also added 16 px of padding to the main content 😬 yeah...probably not what we want.

## Fix
So I moved it to be a parameter that you can pass to the Toast itself and then we set it on _that_ view when we compose the toast. 

That is

instead of setting it here:

```
public func body(content: Content) -> some View {
   content
        .overlay(
            alignment: edge.asAlignment,
            content: warpToastOverlay
        )
        .padding(.horizontal, 16) // <---- here is no good
        .zIndex(1)
        .animation(.default, value: isPresented)
}
```

We set it here

```
@ViewBuilder
func warpToastOverlay() -> some View {
    if isPresented {
        Warp.Toast(
            style: style,
            title: title,
            toastEdge: edge,
            isPresented: $isPresented
        )
        .padding(.horizontal, horizontalPadding) // <---- here is good 👌 
    }
}
```

## Example View
I updated the `ToastView` so you can see how the `horizontalPadding` effects your view

### DIFF
--- a/Sources/Toast/ToastViewModifier.swift
+++ b/Sources/Toast/ToastViewModifier.swift
@@ -5,6 +5,7 @@ extension Warp {
         let style: Warp.ToastStyle
         let title: String
         let edge: ToastEdge
+        let horizontalPadding: Double
         @Binding var isPresented: Bool
 
         public func body(content: Content) -> some View {
@@ -13,7 +14,6 @@ extension Warp {
                     alignment: edge.asAlignment,
                     content: warpToastOverlay
                 )
-                .padding(.horizontal, 16)
                 .zIndex(1)
                 .animation(.default, value: isPresented)
         }
@@ -27,6 +27,7 @@ extension Warp {
                     toastEdge: edge,
                     isPresented: $isPresented
                 )
+                .padding(.horizontal, horizontalPadding)
             }
         }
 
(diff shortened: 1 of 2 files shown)


Sources/Toast/ToastViewModifier.swift
```diff
@@ -38,13 +39,15 @@ public extension View {
         style: Warp.ToastStyle,
         title: String,
         edge: Warp.ToastEdge,
+        horizontalPadding: Double = 0,
```
mohsen-schibsted, line 42: I don't see any requirements from Warp to be able to set the `horizontalPadding`. Is that something your designers asked for? So I can report back to the team that this is something that our users need so we can document it and other platforms can add it too. And should the default value be 16 so by default it has a padding? or 0 is okey?

mohsen-schibsted (approved):

Sources/Toast/ToastViewModifier.swift
mohsen-schibsted: I have asked our designers for feedback regarding padding rules on this component

pbodsk: I have this design.

Toast_example

But maybe we should change the default to 16 and lock it at that...just so your toasts don't go wall to wall but you can't tweak them yourself?

mohsen-schibsted: We have a response from our designers. It should be always constant and 16 from the sides.

pbodsk: Cool, I've updated already and locked it so it can't be modified. I'll merge

PR #23, opened 2024-02-07
Add pill component

mohsen-schibsted:


Sources/Pill/Pill.swift
```diff
@@ -0,0 +1,110 @@
...
+            lhs.text == rhs.text &&
+            lhs.isClosable == rhs.isClosable &&
+            lhs.style == rhs.style
+        }
+        
+        public func hash(into hasher: inout Hasher) {
+            hasher.combine(text)
+            hasher.combine(isClosable)
+            hasher.combine(style)
+        }
```
vishkaischibsted, line 16: No need to check `isClosable`, also there is no need to feed it hasher too.

Sources/Pill/PillStyle+Extension.swift
```diff
@@ -0,0 +1,125 @@
...
+    }
+}
+
+/// Syntactic sugar to ease using warp button style.
+extension ButtonStyle where Self == WarpPillStyle {
+    /// Button style that will transform pill background view to `Warp` design style view.
+    static func warp(
+        style: Warp.PillStyle,
+        colorProvider: ColorProvider
+    ) -> Self {
```
vishkaischibsted, line 100: Can you please change this style name to something else?
Its a bit confusing since it is overlapping with `Warp.Button` style.

Proposed: `warpPill`.

P.S: I don't think changing argument from `style` to `pillStyle` is a good solution, so I didn't mention it.

Sources/Pill/Pill.swift
```diff
@@ -0,0 +1,110 @@
...
+                .warp(
+                    style: style,
+                    colorProvider: colorProvider
+                )
+            )
+        }
+        
+        private var isClosable: Bool {
+            onClose != nil
+        }
```
vishkaischibsted, line 74: If you accepted the top discussion related to this computed property, this property definition can aslo be omitted and on call site it would be only a single if let.

Feel free to ignore this, since it's only a suggestion.

Sources/Pill/Pill.swift
```diff
@@ -0,0 +1,110 @@
...
+                onTap()
+            } label: {
+                HStack(spacing: 8) {
+                    Text(text)
+                    if isClosable {
+                        SwiftUI.Button {
+                            onClose?()
+                        } label: {
+                            Image(systemName: "xmark")
+                        }
```
vishkaischibsted, line 56: When a pill is closable,
Shouldn't we handle the updates?
Or is it the callee's responsibility?!

- If it's our responsibility:
We have to add a boolean state property, when close button is tapped, we will update that boolean property and return nothing as this view's body, in that way pill is completely removed from the view hierarchy.

vishkaischibsted (approved):

Sources/Pill/PillStyle+Extension.swift
mohsen-schibsted: Good idea, then probably we have to change the other one too. Will do that later

Sources/Pill/Pill.swift
mohsen-schibsted: Probably the caller, we still don't know how the users will use these, so probably will do that later on when we have more concrete information how our users will use this.

PR #25, opened 2024-02-09
Box view

vishkaischibsted:
Add `Box` component.

Screenshot 2024-02-09 at 13 27 55


Sources/Box/Box.swift
```diff
@@ -0,0 +1,363 @@
...
+            hasher.combine(title)
+        }
+
+        public init(
+            style: Warp.BoxStyle,
+            title: String?,
+            subtitle: String,
+            link: ButtonConstructor? = nil,
+            button: ButtonConstructor? = nil,
+            secondaryButton: ButtonConstructor? = nil,
```
mohsen-schibsted, line 49: box only has one button, and maybe add the missing self so it looks the same

vishkaischibsted: Done.

Sources/Box/Box.swift
```diff
@@ -0,0 +1,362 @@
...
+                informationView
+                    .padding(.leading, 16)
+
+                Spacer()
+            }
+        }
+
+        @ViewBuilder
+        private var toolTipIconView: some View {
+            if true {
```
mohsen-schibsted, line 120: if there is no title this should be hidden too.

vishkaischibsted: Great catch, I will update it.
I will make it injectable.

Sources/Box/Box.swift
```diff
@@ -0,0 +1,362 @@
...
+                buttonsView
+                    .padding(.top, 4)
+                    .padding(.bottom, 8)
+            }
+        }
+
+        @ViewBuilder
+        private var titleView: some View {
+            if let title {
+                Text(title)
```
mohsen-schibsted, line 151: why not use our own Text(title, style: .title3)?

vishkaischibsted: Done, thanks.

Sources/Text/Text.swift
```diff
@@ -30,7 +30,6 @@ extension Warp {
         public var body: some View {
             SwiftUI.Text(text)
                 .font(from: style.asTypography)
-                .foregroundColor(colorProvider.labelText)
```
mohsen-schibsted, line 33: if you remove this then other places that uses our own Warp.Text suddenly lost their colors and became .black, right?

vishkaischibsted: Good point, I will revert this.

mohsen-schibsted (approved):

PR #27, opened 2024-02-12
Refactor box preview

vishkaischibsted:
Refactor `Box` in demo app in order to make creation configuration more dynamic.

Simulator Screenshot - iPhone 15 Pro - 2024-02-12 at 18 01 29

### DIFF
--- a/SampleApp/BoxView.swift
+++ b/SampleApp/BoxView.swift
@@ -19,6 +19,8 @@ struct BoxView: View {
 
     @State private var hasButton = false
 
+    @Environment(\.colorScheme) private var colorScheme
+
     var body: some View {
         Form {
             createBoxView(basedOn: (hasLink, hasButton))
@@ -42,14 +44,8 @@ struct BoxView: View {
             GroupBox(
                 content: {
                     TextField("Write the desired title", text: $boxTitle)
-                        .padding()
-                        .overlay {
-                            RoundedRectangle(cornerRadius: 4)
-                                .stroke(
-                                    Color.white,
-                                    lineWidth: 1
-                                )
-                        }
+                        .padding(.all, 8)
+                        .textFieldDefaultOverlay(basedOn: colorScheme)
                 }, label: {
                     Text("Title")
                 }
@@ -59,6 +55,7 @@ struct BoxView: View {
                 content: {
                     TextField("Write the desired subtitle", text: $boxSubtitle)
                         .padding(.all, 8)
+                        .textFieldDefaultOverlay(basedOn: colorScheme)
                 }, label: {
                     Text("Subtitle")
                 }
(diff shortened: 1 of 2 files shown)


SampleApp/BoxView.swift
```diff
@@ -9,31 +9,155 @@ import SwiftUI
...
+    @State private var hasButton = false
+
+    @Environment(\.colorScheme) private var colorScheme
+
     var body: some View {
-        ScrollView(showsIndicators: false) {
-            ForEach(Warp.BoxStyle.allCases, id: \.self) { style in
-                Warp.Box(
-                    style: style,
-                    title: "Title",
```
mohsen-schibsted, line 17: We don't have this variant, so if title is nil the icon should hide also. Probably this is the change we need to make. Otherwise the demo app looks great.
Simulator Screenshot - iPhone 15 - 2024-02-13 at 19 57 24
Screenshot 2024-02-13 at 20 00 18

mohsen-schibsted: oh no the logic for hiding title when it is empty is in the demoApp :D 
We should move it to the Warp component itself, otherwise it won't work for the users.
Screenshot 2024-02-13 at 20 07 16
Screenshot 2024-02-13 at 20 06 13

SampleApp/BoxView.swift
```diff
@@ -9,31 +9,155 @@ import SwiftUI
...
+    @State private var hasLink = false
+
+    @State private var hasButton = false
+
+    @Environment(\.colorScheme) private var colorScheme
+
     var body: some View {
-        ScrollView(showsIndicators: false) {
-            ForEach(Warp.BoxStyle.allCases, id: \.self) { style in
-                Warp.Box(
```
mohsen-schibsted, line 15: Our components should by default have a padding.
Simulator Screenshot - iPhone 15 - 2024-02-13 at 20 09 05

SampleApp/BoxView.swift
vishkaischibsted: Great catch, thanks!
Updated.

SampleApp/BoxView.swift
vishkaischibsted: Thanks for the finding, you mean should I add paddings to the box itself?

mohsen-schibsted: yes, inside the component we should add padding that users should use, so different users do not add random padding for their cases.

mohsen-schibsted (approved):

SampleApp/BoxView.swift
mohsen-schibsted: Don't worry about it thou if it requires too much changes. We should sync this across all our components probably. The things that should be full screen need to have a padding that we can control and not the user. We have not started working on the padding, our paddings are static now which should not be the case and should be dynamic and a multiple of 4 probably.

PR #29, opened 2024-02-13
Add broadcast view

mohsen-schibsted:
https://www.figma.com/file/nkiRpuVu6XRfvY96BA80H8/Components-overview?type=design&node-id=253-14894&mode=design&t=VZPIm7S3sbphnYGT-0

### DIFF
--- a/Sources/Broadcast/Broadcast.swift
+++ b/Sources/Broadcast/Broadcast.swift
@@ -19,7 +19,7 @@ extension Warp {
     public struct Broadcast: View {
 
         /// Text that will be shown in broadcast
-        let title: String
+        let text: String
 
         /// Edge from where the broadcast is presented
         let broadcastEdge: Warp.BroadcastEdge
@@ -31,29 +31,31 @@ extension Warp {
         let colorProvider: ColorProvider
 
         /**
-         - Parameter title: String to display in the `Broadcast`
+         - Parameter text: String to display in the `Broadcast`
          - Parameter broadcastEdge: The `BroadcastEdge` on where to present the `Broadcast`
          - Parameter isPresented: Is the `Broadcast` presented or not
          - Parameter colorProvider: ColorProvider used for styling the `Broadcast`, default value is read from `Config`
          */
         public init(
-            title: String,
+            text: String,
             broadcastEdge: Warp.BroadcastEdge,
             isPresented: Binding<Bool>,
             colorProvider: ColorProvider = Config.colorProvider
         ) {
-            self.title = title
+            self.text = text
             self.broadcastEdge = broadcastEdge
             self._isPresented = isPresented
             self.colorProvider = colorProvider
         }
 
         public var body: some View {
-            ZStack {
-                RoundedRectangle(cornerRadius: broadcastCornerRadius)
-                    .stroke(colorProvider.broadcastBorderSubtle, lineWidth: 4)
-                contentView
-            }
+            contentView
+            .overlay(
+                HStack(spacing: 0) {
+                    RoundedRectangle(cornerRadius: broadcastCornerRadius)
+                        .stroke(colorProvider.broadcastBorderSubtle, lineWidth: 4)
+                }
+            )
             .frame(maxWidth: .infinity)
             .background(colorProvider.broadcastBackground)
             .cornerRadius(broadcastCornerRadius)
@@ -69,12 +71,14 @@ extension Warp {
             HStack(spacing: 0) {
                 colorProvider.broadcastBorder
                     .frame(width: 4)
-                HStack(spacing: 0) {
-                    Image(systemName: "exclamationmark.triangle.fill")
-                        .renderingMode(.template)
-                        .foregroundColor(colorProvider.broadcastIcon)
+                HStack(alignment: .top, spacing: 0) {
+                    VStack {
+                        Image(systemName: "exclamationmark.triangle.fill")
+                            .renderingMode(.template)
+                            .foregroundColor(colorProvider.broadcastIcon)
+                    }
                     
-                    Text(title, style: .body)
+                    Text(text, style: .body)
                         .padding(.leading, 8)
                         .foregroundStyle(colorProvider.broadcastText)
                     
@@ -87,14 +91,14 @@ extension Warp {
         }
         
         private var leftLineView: some View {
-            return colorProvider.alertWarningIcon
+            return colorProvider.broadcastBorder
         }
     }
 }
 
 #Preview {
     Warp.Broadcast(
-        title: "This is a success broadcast",
+        text: "This is a success broadcast",
         broadcastEdge: .top,
         isPresented: .constant(true)
     )
--- a/Sources/Broadcast/Broadcast.swift
+++ b/Sources/Broadcast/Broadcast.swift
@@ -1,0 +1,97 @@
+import Foundation
+import SwiftUI
+
+extension Warp {
+    private static let broadcastCornerRadius: Double = 8
+
+    /**
+        Longer broadcast message that spans over several lines. Displayed at the top of the page or contextually within a task flow.
+
+        **When to use**
+
+        Use to make an announcement or provide information outside of the normal user flow.
+
+         To use the iOS `Broadcast` View you need to provide it with
+        - a `title`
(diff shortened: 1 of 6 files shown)


Sources/Broadcast/Broadcast.swift
```diff
@@ -0,0 +1,105 @@
...
+        public var body: some View {
+            contentView
+            .overlay(
+                ZStack(alignment: .leading) {
+                    RoundedRectangle(cornerRadius: broadcastCornerRadius)
+                        .stroke(colorProvider.broadcastBorderSubtle, lineWidth: 4)
+                    colorProvider.broadcastBorder
+                        .frame(width: 6)
+                }
+            )
```
vishkaischibsted, line 60: Missed one tab indentation!

Sources/Broadcast/Broadcast.swift
```diff
@@ -0,0 +1,105 @@
...
+            .frame(maxWidth: .infinity)
+            .background(colorProvider.broadcastBackground)
+            .cornerRadius(broadcastCornerRadius)
+            .transition(AnyTransition.move(edge: broadcastEdge.asEdge).combined(with: .opacity))
```
vishkaischibsted, line 64: You can omit `AnyTransition`.

Feel free to ignore this, since it's only a suggestion.

Sources/Broadcast/Broadcast.swift
```diff
@@ -0,0 +1,105 @@
...
+            .onTapGesture {
+                withAnimation {
+                    isPresented.toggle()
```
vishkaischibsted, line 67: Flag should always be set to `false`.

Sources/Broadcast/Broadcast.swift
```diff
@@ -0,0 +1,105 @@
...
+                }
+            }
+        }
+
+        private var contentView: some View {
+            HStack(spacing: 0) {
+                HStack(alignment: .top, spacing: 0) {
```
vishkaischibsted, line 74: Was it necessary to have two `HStack`?

Sources/Broadcast/Broadcast.swift
```diff
@@ -0,0 +1,105 @@
...
+                    VStack {
+                        Image(systemName: "exclamationmark.triangle.fill")
+                            .renderingMode(.template)
+                            .foregroundColor(colorProvider.broadcastIcon)
+                    }
```
vishkaischibsted, line 79: Same here, `VStack` can be omitted,
Also `Image` can be extracted as a computed property.
Proposing:  `warningImageView`

Sources/Broadcast/Broadcast.swift
```diff
@@ -0,0 +1,105 @@
...
+                    
+                    Text(text, style: .body)
+                        .padding(.leading, 8)
+                        .foregroundStyle(colorProvider.broadcastText)
+                    
+                    Spacer()
+                    
+                    Image("icon-close", bundle: .module)
```
vishkaischibsted, line 87: Same here,
`Text` and `Image` can be extracted to a new computed properties.

Sources/Broadcast/BroadcastEdge.swift
```diff
@@ -0,0 +1,29 @@
...
+
+extension Warp {
+    /**
+        Enum describing  which edge the Broadcast view will appear from/disappear to
+     */
+    public enum BroadcastEdge: CaseIterable {
+        case top
+        case bottom
+
+        public var asAlignment: Alignment {
```
vishkaischibsted, line 11: Should we have this property as public?
Also when we have it inside the domain itself, we are glueing ui and logic together,
Same as next computed property `asEdge`.

Sources/Broadcast/BroadcastViewModifier.swift
```diff
@@ -0,0 +1,49 @@
+import SwiftUI
+
+extension Warp {
+    public struct BroadcastViewModifier: ViewModifier {
```
vishkaischibsted, line 4: Struct is public but initializer/constructor is not accessible from outside.

Sources/Broadcast/BroadcastViewModifier.swift
```diff
@@ -0,0 +1,49 @@
...
+        let text: String
+        let edge: BroadcastEdge
+        private let horizontalPadding: Double = 16
```
vishkaischibsted, line 7: Is this a static constant?
If not then it can not be injected now.

Sources/Broadcast/BroadcastViewModifier.swift
```diff
@@ -0,0 +1,49 @@
...
+                .overlay(
+                    alignment: edge.asAlignment,
+                    content: warpBroadcastOverlay
+                )
+                .zIndex(1)
+                .animation(.default, value: isPresented)
+        }
+
+        @ViewBuilder
+        func warpBroadcastOverlay() -> some View {
```
vishkaischibsted, line 21: This should be marked as private.

Sources/Broadcast/BroadcastViewModifier.swift
```diff
@@ -0,0 +1,49 @@
...
+    }
+}
+
+public extension View {
+    func warpBroadcast(
+        text: String,
+        edge: Warp.BroadcastEdge,
+        isPresented: Binding<Bool>
+    ) -> some View {
+        self.modifier(
```
vishkaischibsted, line 41: self can be omitted.

Sources/Broadcast/Broadcast.swift
```diff
@@ -0,0 +1,105 @@
...
+            text: String,
+            broadcastEdge: Warp.BroadcastEdge,
+            isPresented: Binding<Bool>,
+            colorProvider: ColorProvider = Config.colorProvider
+        ) {
+            self.text = text
+            self.broadcastEdge = broadcastEdge
+            self._isPresented = isPresented
+            self.colorProvider = colorProvider
+        }
```
vishkaischibsted, line 49: In one hand init can be optional, as we can deal with the `isPresented` as soon as the view wants to be created,
But in the other hand this implementation delays handing empty state during the view lifecycle.

Either way looks good, I wanted to point out the other possibility.
Feel free to ignore this comment since it's only a suggestion!

Sources/Broadcast/Broadcast.swift
mohsen-schibsted: Don't know :D Let's check.

mohsen-schibsted: nope, removed it

Sources/Broadcast/BroadcastEdge.swift
mohsen-schibsted: Since I copy and pasted this from Toast which @pbodsk wrote I am not gonna play with it now. We can discuss this later and see what's best.

Sources/Broadcast/BroadcastViewModifier.swift
mohsen-schibsted: It is working in the DemoApp so users should be fine. Again copy from Toast so if we want to fix something we can do both on a later branch.

Sources/Broadcast/BroadcastViewModifier.swift
mohsen-schibsted: Yes I would say padding we control and users should not be able to change it.

Sources/Broadcast/Broadcast.swift
mohsen-schibsted: Since this is copy from Toast we will deal with improvements later.

Sources/Broadcast/BroadcastEdge.swift
```diff
@@ -0,0 +1,29 @@
+import SwiftUI
+
+extension Warp {
+    /**
+        Enum describing  which edge the Broadcast view will appear from/disappear to
+     */
+    public enum BroadcastEdge: CaseIterable {
```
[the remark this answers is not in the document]
mohsen-schibsted: Since it will break stuff I will leave `CaseIterable` as is, we can discuss what should be public or not later, we have that in many places.

vishkaischibsted (approved):

PR #30, opened 2024-02-15
Refactor button preview

vishkaischibsted:
Refactor `ButtonView` in demo app in order to make creation configuration dynamic.

Simulator Screenshot - iPhone 15 Pro - 2024-02-15 at 13 59 28

Simulator Screenshot - iPhone 15 Pro - 2024-02-15 at 13 59 48

### DIFF
--- a/SampleApp/ButtonView.swift
+++ b/SampleApp/ButtonView.swift
@@ -1,6 +1,19 @@
 import SwiftUI
 import Warp
 
+    private func createButton() -> some View {
+        Warp.Button.create(
+            for: type,
+            title: buttonTitle,
+            icon: buttonHasIcon ? Image(systemName: "plus"): nil,
+            action: {},
+            size: buttonSize,
+            isEnabled: isButtonEnabled,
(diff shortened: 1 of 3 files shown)


SampleApp/ButtonView.swift
```diff
@@ -1,6 +1,126 @@
...
+                    .pickerStyle(.wheel)
+                    .defaultPadding()
+                }, label: {
+                    Text("Type")
+                }
+            )
+
+            GroupBox(
+                content: {
+                    TextField("Write the desired title", text: $buttonTitle)
```
mohsen-schibsted, line 41: If we replace this with `Warp.TextField` will the `colorScheme` be removed? We should use our own components as much as we can, even in the `DemoApp`

mohsen-schibsted (approved):

SampleApp/ButtonView.swift
vishkaischibsted: Good point, updated 👍

PR #35, opened 2024-03-01
Add shadow

mohsen-schibsted:
https://www.figma.com/file/oHBCzDdJxHQ6fmFLYWUltf/Warp---Components-2.0?type=design&node-id=6165-5119&mode=design&t=YYKVaK0zAD6nR29q-0

### DIFF
--- a/SampleApp/ShadowView.swift
+++ b/SampleApp/ShadowView.swift
@@ -31,7 +31,7 @@ struct ShadowView: View {
                 Spacer()
                 Picker("Pick your shadow", selection: $shadow.animation(.interpolatingSpring)) {
                     ForEach(Warp.Shadow.allCases, id: \.self) { shadow in
-                        Text(shadow.rawValue)
+                        Text(shadow.name)
                     }
                 }
                 .pickerStyle(.segmented)
@@ -50,7 +50,7 @@ extension Warp.Shadow {
         .xLarge
     ]
     
-    fileprivate var rawValue: String {
+    fileprivate var name: String {
         switch self {
             case .small:
                 return "Small"
--- a/SampleApp/ShadowView.swift
+++ b/SampleApp/ShadowView.swift
@@ -3,7 +3,7 @@ import SwiftUI
 
 struct ShadowView: View {
     private let colorProvider: ColorProvider = Warp.Config.colorProvider
-    @State private var shadow = Shadow.small
+    @State private var shadow = Warp.Shadow.small
 
     var body: some View {
         ZStack {
@@ -30,7 +30,7 @@ struct ShadowView: View {
                 .shadow(shadow)
                 Spacer()
                 Picker("Pick your shadow", selection: $shadow.animation(.interpolatingSpring)) {
-                    ForEach(Shadow.allCases, id: \.self) { shadow in
+                    ForEach(Warp.Shadow.allCases, id: \.self) { shadow in
                         Text(shadow.rawValue)
                     }
                 }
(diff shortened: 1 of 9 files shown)


Sources/Button/WarpButtonStyle/UIMetricsFactory.swift
```diff
@@ -56,13 +41,10 @@ extension Warp.Button {
...
         var cornerRadius: CGFloat {
-            switch type {
-                case .utility:
-                    return 4
-
-                default:
-                    return 8
+            if type.isUtilityRelatedButton {
+                return 4
             }
```
vishkaischibsted, line 46: Please add an empty line after "}".

Sources/Button/WarpButtonStyle/UIMetricsFactory.swift
```diff
@@ -18,29 +18,14 @@ extension Warp.Button {
...
-            switch type {
-                case .utilityOverlay:
-                    return isEnabled ? 1: 0
-
-                default:
-                    return 0
+        var shadow: Shadow? {
+            if type == .utilityOverlay, isEnabled {
+                return .small
             }
```
vishkaischibsted, line 27: Please add an empty line after "}".

Sources/Button/WarpButtonStyle/UIMetricsFactory.swift
```diff
@@ -18,29 +18,14 @@ extension Warp.Button {
             self.size = size
             self.isEnabled = isEnabled
         }
-
-        /// Button shadow radius.
+        
+        /// Button shadow
```
vishkaischibsted, line 22: Add "." at the end.

Sources/Shadow/Shadow.swift
```diff
@@ -0,0 +1,106 @@
+import SwiftUI
+
+public enum Shadow: String, CaseIterable {
```
vishkaischibsted, line 3: Remove `String` conformance, and based on your design remove `CaseIterable` if possible.

Sources/Shadow/Shadow.swift
```diff
@@ -0,0 +1,106 @@
...
+    case small, medium, large, xLarge
+    
+    var opacity1: Double {
```
vishkaischibsted, line 6: Mark all the computed properties as private please.

Sources/Shadow/Shadow.swift
```diff
@@ -0,0 +1,106 @@
...
+            return 3
+        case .large:
+            return 6
+        case .xLarge:
+            return 9
+        }
+    }
+}
+
+struct ShadowType: ViewModifier {
```
vishkaischibsted, line 86: If this should be expose to user,
Mark it as public,
But if this should not be exposed, then mark it as private in order to hide detail implementation from outside of this file.

Also rename it to something more meaningful,
Proposed: ShadowViewModifier

Sources/Shadow/Shadow.swift
```diff
@@ -0,0 +1,106 @@
...
+    var shadow: Shadow
```
vishkaischibsted, line 87: Make it a constant please.

Sources/Shadow/Shadow.swift
mohsen-schibsted: I need the `String` for the title of the `picker` on the `demoApp`. Should I make a variable in the `DemoApp` then?

vishkaischibsted: @mohsen-schibsted Exactly, lets encapsulate the logics related to `DemoApp` inside of it.
The same as `CaseIterable`, who needs to add all the shadows to a single view in consumer side.

SampleApp/ShadowView.swift
```diff
@@ -0,0 +1,72 @@
...
+
+extension Warp.Shadow {
+    fileprivate static var allCases: [Warp.Shadow] = [
+        .small,
+        .medium,
+        .large,
+        .xLarge
+    ]
+    
+    fileprivate var rawValue: String {
```
vishkaischibsted, line 53: `rawValue` is a bit vague,
`name` or `title` would be more appropriate.

Feel free to ignore this comment.

vishkaischibsted (approved):

PR #44, opened 2024-06-05
Bugfix/alertview truncated text

pbodsk:
# Bugfix Suggestion for Alert

I noticed that our `Alert` component would truncate text if the text was sufficiently long...but only in *Tori*.

Here's now that looks (notice the "We apologize for t...")

Tori existing

So I took a look to see if I could fix it and I think I've got it now

### AlertView
The main attraction is of course the `Alert` view itself. I made the following changes:

- got rid of the `ZStack` in the `body` as I find they often mess with the layout. Instead the background is now placed in a `.background`
- shuffled things around in the `foregroundView` 
  - moved the `leftLineView` to an `.overlay`
  - made icon and text top align in the `HStack`
  - added padding to the `HStack` so we control it "globally" for that view
- since we control padding globally for the `HStack` and since we top align icon and text I could get rid of a lot of the padding in the `toolTipIconView`
- added `.fixedSize(horizontal: false, vertical: true)` to the `informationView` as that seemed to be the only way to ensure that text was not truncated when you started adding links and buttons to the mix 🤔 
- the preview builder was complaining about properties on `Text` that it couldn't find...because we were actually using a `Warp.Text`, so I prefixed the `Text` used in `titleView` and `subtitleView` and then preview worked again...just makes it so much easier to experiment 😅

And here is how it looks now with the same text in Tori

Tori fix

### Removing No Longer Building Tests
Some of the tests were no longer building and was using something that seems to have been gone since version 0.0.6 so...they were deleted as well, don't know if we need them 🤷 .

## And
That's it 😅 

Hope it is useful 🤞


mohsen-schibsted:
Thank you @pbodsk for contributing to WARP 🥳 yeah I remember seeing this issue before (not in WARP) and it behaved differently depending on the accessibility setting, custom font and device. I guess the system does not calculate the height of the label correctly.

This does not happen to me using your text, what device and version did you use?

And comparing to main branch the attached video happens for me, Tori Alert is a bit smaller in width, but Finn Alert is a lot smaller. Any ideas why that happens for me and not for you?


https://github.com/warp-ds/warp-ios/assets/89784012/0b65aeba-6b38-4457-ab86-f079f8d2ace4

pbodsk:
Hey @mohsen-schibsted, welcome back
> This does not happen to me using your text, what device and version did you use?

I use an iPhone 15 Pro with iOS 17.5 and this text:

> Due to regularory requirements, we need to collect your social security nummer. Therefore please verify your information again. We apologize for the inconvenience

And here's how that looks on the `main` branch:

Simulator Screenshot - iPhone 15 Pro - 2024-06-24 at 13 59 39

pbodsk:
> And comparing to main branch the attached video happens for me, Tori Alert is a bit smaller in width, but Finn Alert is a lot smaller. Any ideas why that happens for me and not for you?

I think I do 😅 

As far as I recall I no longer has a `.frame(maxWidth: .infinity)` on the alert component, so it is dynamic in size, compared to the existing that always take up all space.

I can fix that (I think!) so we always take up the entire width (minus padding of course). Let me take  a look! 😅

pbodsk:
@mohsen-schibsted There we go, the alert is now full width 🤞

mohsen-schibsted (approved):

mohsen-schibsted:
> @mohsen-schibsted There we go, the alert is now full width 🤞

Works great now 🥳 thank you

PR #46, opened 2024-06-21
StepIndicator component

pbodsk:
I had some of the code for a `StepIndicator` laying from the KYC feature so I've `Warp`-ified that and added the missing features.

## How Does it Look? 🔍 
Here are a couple of animated gifs showing you how this looks in the UI now. As you can see I've made a small sandbox in the demo app where you can experiment with the different states...that is probably not the use case in the real world but...just so you can see how the different states look

|Horizontal|Vertical|
|---|---|
|Horizontal|Vertical|

## Reading Order 👀 
Here is a suggested reading order for the files in this PR

### `StepIndicator`
A good first candidate is the `StepIndicator` itself, just to get an idea about what goes on here.

You'll notice that the `StepIndicator` takes two - somewhat interesting - parameters.

- A `LayoutOrientation` indicating if you want to use a `.horizontal` or a `.vertical` `StepIndicator`.
- An array of `StepIndicatorItem`s which are simple `struct`s for the individual items in a `StepIndicator`.

Here's the `.init` for the `StepIndicator`:

```swift
public init(
    layoutOrientation: LayoutOrientation = .vertical,
    steps: [StepIndicatorItem],
    colorProvider: ColorProvider = Config.colorProvider
) {
    self.layoutOrientation = layoutOrientation
    self.orderedSteps = steps.compactMap { step in
        OrderedStepIndicatorItem(item: step, items: steps)
    }
    self.colorProvider = colorProvider
}
```

Note the `.compactMap` to a `OrderedStepIndicatorItem`...what is that?

#### `OrderedStepIndicatorItem`
We need to use an `internal` `OrderedStepIndicatorItem` because in order to draw the lines between the individual steps, we need to know what comes before and after the current step.

This is not a concern for the end consumer of this API. As long as they provide us with a sorted array of `StepIndicatorItem`s, we'll handle the logic related to drawing.

The `.init` of `OrderedStepIndicatorItem` will fetch the progress of the (potential) steps before and after the current step and store these as associated values to `StepIndicatorItem.Position` enum values that can be used when drawing the lines between steps.

I've written a bunch of tests for this in `OrderedStepIndicatorItemTests` to help prove the correctness of the logic in the `init` method, but I agree that it is _a bit hairy_ 😬 

### Next stop - a quick visit to the `body` of `StepIndicator`
If you look at the `body` of `StepIndicator` itself you'll see that it switches over the `LayoutOrientation` to determine whether to `ForEach` over a list of  `HorizontalItemView` or `VerticalItemView` views.

(the horizontal variant wraps its content inside a horizontal `ScrollView` which I deemed to be the required behaviour out of the box (so the consumer doesn't have to wrap this themselves))

###  `HorizontalItemView` and `VerticalItemView`
Not a lot of exiting stuff going on here, most of the files are `#Preview`s to ensure I got all tracks covered and the rest is aggregating text from the `DescriptionView` and progress shapes from the `HorizontalProgressView/VerticalProgressView`

### `HorizontalProgressView` and `VerticalProgressView`
This is where we draw the lines and circles needed to illustrate the progress.

They differ a bit in the way they are drawn.

#### `HorizontalProgressView`
Since the circle needs to be centered with the text above it, we draw horizontal progress like so:

|First Item|Second Item|Third Item
|---|---|---|
|EmptyView-Circle-Rectangle|Rectangle-Circle-Rectangle|Rectangle-Circle-EmptyView

That means that the line between the first and second view is "shared" between them, so each draws a half.

ℹ️ Going back to the `OrderedStepIndicatorItem`,  _this_  is why the steps needs to know about the progress of views before and after them. ℹ️ 

This is most visible in the vars and functions:

- `HorizontalProgressView.getter:leadingView`
- `HorizontalProgressView.getter:trailingView`
- `HorizontalProgressView.line(for:ownProgress:)`

Where you'll find the logic for this drawing.

#### `VerticalProgressView`
Vertical views are easier to draw. Here we just need to draw a circle with - potentially - a rectangle under it, so we define a custom `VerticalProgressShape` for that purpose and draw and fill that.

And that concludes the tour of the Step Indicator feature.

### Further Additions
I've taken the liberty of cleaning up our demo app so the items now appear in two sections:

- Brand Specific Items (colors etc)
- Warp Components (now without the Warp prefix and sorted alphabetically)

Simulator Screenshot - iPhone 15 Pro - 2024-06-21 at 11 55 23

## Parting Words
Thank you for reading all the way to the end 😅 I hope this description helps with clarity over this feature 🤞


SampleApp/Warp-iosFinn/ContentView.swift
```diff
@@ -20,91 +20,130 @@ struct ContentView: View {
                     Config.warpTheme = selectedTheme
                 }
                 #endif
-                VStack (alignment: .leading) {
+                LazyVStack (alignment: .leading) {
```
mohsen-schibsted, line 23: Thank you for the clean up 🥳

Sources/Components/StepIndicator/ViewHelpers/LineBuilder.swift
```diff
@@ -0,0 +1,35 @@
...
+            for otherProgress: Warp.StepIndicatorItem.Progress,
+            ownProgress: Warp.StepIndicatorItem.Progress,
+            colorProvider: ColorProvider = Warp.Config.colorProvider,
+            orientation: Warp.StepIndicator.LayoutOrientation,
+            lineThickness: Double = 2
+        ) -> some View {
+
+            let fillColor = switch otherProgress {
+            case .incomplete:
+                colorProvider.token.backgroundDisabled
```
mohsen-schibsted, line 16: Please use the `stepIndicatorColors` in `colorProvider` directly, for example this one can be `stepIndicatorTrackBackground`

Sources/Components/StepIndicator/Subviews/VerticalProgressView.swift
```diff
@@ -0,0 +1,122 @@
...
+                case .inProgress:
+                    Circle()
+                        .fill(progress.fillColor(using: colorProvider))
+                case .complete:
+                    Circle()
+                        .fill(progress.fillColor(using: colorProvider))
+                        .overlay(alignment: .leading) {
+                            Image("icon-stepindicator-completed", bundle: .module)
+                                .renderingMode(.template)
+                                .foregroundColor(colorProvider.token.iconInverted)
```
mohsen-schibsted, line 39: `stepIndicatorHandleIcon`

pbodsk: And updated here too

mohsen-schibsted (approved):

PR #48, opened 2024-06-25
Modal component

milad-schibsted:
https://www.figma.com/design/oHBCzDdJxHQ6fmFLYWUltf/WARP---Components-2.0?node-id=1308-38520&t=LOcFe01WHgUwGKy7-0


Sources/Components/Modal/Modal.swift
```diff
@@ -0,0 +1,237 @@
...
+        private var subtitleView: some View {
+            if let subtitle, !subtitle.isEmpty {
+                Text(subtitle, style: .title4)
+                    .accessibilityRemoveTraits(.isHeader)
+            }
+        }
+        
+        private var bodyTextView: some View {
+            Text(bodyText, style: .body)
+                .accessibilityRemoveTraits(.isHeader)
```
pbodsk, line 109:
```suggestion
            Text(bodyText, style: .body)
                .fixedSize(horizontal: false, vertical: true)
                .accessibilityRemoveTraits(.isHeader)
```

pbodsk: I noticed that the modal clips text on an iPhone 13 mini as you can see here:

Simulator Screenshot - iPhone 13 mini - 2024-06-25 at 11 19 48

I guess the error is similar to the one I've attempted to fix in #44 where the solution was to introduce a `.fixedSize` modifier like I've suggested above.

You should probably check if that introduces any side effects elsewhere but I just wanted to suggest a - potential - fix instead of just dumping an error on you 😄

pbodsk (commented):
Hey @milad-schibsted 


A quick observation and a potential fix.

Oh...and also, you need to add the `ModalView` to the tori and DBA targets. I had problems getting them to build before I added it.

Screenshot 2024-06-25 at 12 51 38

Sources/Components/Modal/Modal.swift
milad-schibsted: Yes, `.fixedSize(horizontal: false, vertical: true)` seems fix the issue.

SampleApp/ModalView.swift
```diff
@@ -0,0 +1,202 @@
...
+            primaryButton: getPrimaryButtonProvider(),
+            secondaryButton: getSecondaryButtonProvider(),
+            closeButton: hasCloseButton,
+            dismissOnClickOutside: dismissOnClickOutside,
+            isPresented: $presentModal,
+            colorProvider: Warp.Config.colorProvider)
+    }
+
+    private func createModalView(
+        basedOn state: (hasPrimaryButton: Bool, hasSecondaryButton: Bool, hasCloseButton: Bool)
```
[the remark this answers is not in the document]
milad-schibsted: It is used to create this preview for the modal.

Sources/Components/Modal/Modal.swift
```diff
@@ -0,0 +1,238 @@
...
+        /// Tuple that will provide a title and an action for creating a primary button view.
+        /// Passing `nil` will skip adding primary button view.
+        let primaryButtonProvider: ButtonConstructor?
+        
+        /// Tuple that will provide a title and an action for creating a secondary button view.
+        /// Passing `nil` will skip adding secondary button view.
+        let secondaryButtonProvider: ButtonConstructor?
+        
+        /// Show/Hide close  image view.
+        var closeButton: Bool
```
mohsen-schibsted, line 34: Maybe hasCloseButton? And it seems Android has an `onDismiss` action too.

Sources/Components/Modal/ModalViewModifier.swift
```diff
@@ -0,0 +1,157 @@
...
+        let primaryButtonProvider: ButtonConstructor?
+        
+        
+        
+        /// Tuple that will provide a title and an action for creating a secondary button view.
+        /// Passing `nil` will skip adding secondary button view.
+        let secondaryButtonProvider: ButtonConstructor?
+        
+        /// boolean show/hide close image view.
+        var closeButton: Bool
```
mohsen-schibsted, line 37: Same as above, hasCloseButton? or showCloseButton?

Sources/Components/Modal/Modal.swift
```diff
@@ -0,0 +1,246 @@
...
+        }
+        
+        @ViewBuilder
+        private var closeButtonView: some View {
+            if hasCloseButton {
+                Image("icon-close", bundle: .module)
+                    .onTapGesture {
+                        isPresented.toggle()
+                        guard let onDismiss = onDismiss else { return }
+                        onDismiss()
```
mohsen-schibsted, line 102: would just having `ondismiss?()` do the same thing?

Sources/Components/Modal/ModalViewModifier.swift
```diff
@@ -0,0 +1,168 @@
...
+        
+        @ViewBuilder
+        func warpModalOverlay() -> some View {
+            if isPresented {
+                ZStack {
+                    Color.black.opacity(0.4)
+                        .onTapGesture {
+                            guard dismissOnClickOutside else { return }
+                            isPresented.toggle()
+                            guard let onDismiss = onDismiss else { return }
```
mohsen-schibsted, line 105: would just having `ondismiss?()` do the same thing?

SampleApp/ModalView.swift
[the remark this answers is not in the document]
milad-schibsted: Ah, I see 😅. You're right 
I'll remove it. :)

mohsen-schibsted (approved):

PR #49, opened 2024-06-27
Adding Spacing and Border

dinsen:
Introducing `Spacing` and `Border` 🎉 

With the new enums we are now able to specify values for padding and sizes like: `Warp.Spacing.spacing100` and `Warp.Border.borderRadius50`.

I know on Android they choose names like `space025`, `space1`, `space8`. 
But after some chatting with @pbodsk and Christian Therkelsen, we choose to go with `spacing25`, `spacing100`, `spacing800`, since it's easier to understand what the name represent. 

`spacing25` = 2 (8 / 4)
`spacing50` = 4 (8 / 2)
`spacing100` = 8 <-- baseline
`spacing200` = 16 (8 * 2)
`spacing250` = 20 (8 * 2.5)
`spacing300` = 20 (8 * 3)

We also talked about having all the values in `Warp.Dimensions` but we already have `Warp.Shadow` and `Warp.Typography` so it might make sense to have `Warp.Spacing` and `Warp.Border` as well.

But everything is up for discussion 📣

### DIFF
--- a/Sources/Spacing/Spacing.swift
+++ b/Sources/Spacing/Spacing.swift
@@ -1,0 +1,42 @@
+import CoreFoundation
+
+extension Warp {
+    public enum Spacing: CGFloat {
+        /// Separation of 2 points.
+        case spacing25 = 2
+        /// Separation of 4 points.
+        case spacing50 = 4
+        /// Separation of 8 points.
+        case spacing100 = 8
+        /// Separation of 12 points.
+        case spacing150 = 12
+        /// Separation of 16 points.
+        case spacing200 = 16
+        /// Separation of 20 points.
+        case spacing250 = 20
+        /// Separation of 24 points.
+        case spacing300 = 24
+        /// Separation of 32 points.
+        case spacing400 = 32
+        /// Separation of 40 points.
+        case spacing500 = 40
+        /// Separation of 48 points.
+        case spacing600 = 48
+        /// Separation of 56 points.
+        case spacing700 = 56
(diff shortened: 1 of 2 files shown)


Sources/Spacing/Spacing.swift
```diff
@@ -0,0 +1,42 @@
+import CoreFoundation
+
+extension Warp {
+    public enum Spacing: CGFloat {
+        /// Separation of 2 points.
+        case spacing25 = 2
```
mohsen-schibsted, line 6: Just a heads up that these does not exist on WARP as of yet so I will deprecate these when we have a source of truth on Figma.
But as long as we don't have that feel free to add anything you need with the structure that makes sense to you here 👍

mohsen-schibsted (approved):

Sources/Spacing/Spacing.swift
dinsen: I actually just added it because android have it 🙈

mohsen-schibsted: 😅👍 It should be great, now at least we have a place and structure for it 🥳 we will fill them with correct numbers when WARP is ready.

dinsen:
@mohsen-schibsted I don't have write access, so you can merge when you have time 🚀

mohsen-schibsted:
> @mohsen-schibsted I don't have write access, so you can merge when you have time 🚀

Now you should have that after accepting the invite 🥳 Thank you again for all the contributions

PR #50, opened 2024-06-27
Adding Token and UIToken variables to make "shortcuts" for TokenProvider and UITokenProvider

dinsen:
I think we talked about it before @mohsen-schibsted. That it would be great if developers only needed to type `Warp.Token.text` instead of `Warp.Config.tokenProvider.text`.

This PR is a suggestion to make new public variables that the developers should use, instead of using the variables from `Config`.

Later we should remove the public keyword from the variables in `Config`. 
Since this is a breaking change, I haven't removed the public keyword. 

Let's have a discussion if we should call it something different? 
And what about colors. Should they also be public available?

### DIFF
--- a/Sources/Config.swift
+++ b/Sources/Config.swift
@@ -1,6 +1,14 @@
 extension Warp {
+    public static var Token: TokenProvider {
+        return Config.tokenProvider
+    }
+
+    public static var UIToken: UITokenProvider {
+        return Config.uiTokenProvider
+    }
+
     public struct Config {
-        public static var tokenProvider: TokenProvider = {
+        static var tokenProvider: TokenProvider = {
             switch warpTheme {
             case .finn:
                 return FinnTokenProvider()
@@ -10,7 +18,8 @@ extension Warp {
                 return DBATokenProvider()
             }
         }()
-        public static var uiTokenProvider: UITokenProvider = {
+
+        static var uiTokenProvider: UITokenProvider = {
             switch warpTheme {
             case .finn:
                 return FinnUITokenProvider()
--- a/Sources/Config.swift
+++ b/Sources/Config.swift
@@ -8,7 +8,7 @@ extension Warp {
     }
 
     public struct Config {
-        static var tokenProvider: TokenProvider = {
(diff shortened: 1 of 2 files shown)


mohsen-schibsted:
> I think we talked about it before @mohsen-schibsted. That it would be great if developers only needed to type `Warp.Token.text` instead of `Warp.Config.tokenProvider.text`.
> 
> This PR is a suggestion to make new public variables that the developers should use, instead of using the variables from `Config`.
> 
> Later we should remove the public keyword from the variables in `Config`. Since this is a breaking change, I haven't removed the public keyword.
> 
> Let's have a discussion if we should call it something different? And what about colors. Should they also be public available?

Thank you, this looks great 🥳 Generally no people should not have access to Colors and should use Tokens. Colors are for internal use in WARP components. But since we have cases that we had to sacrifice and use WARP colors in some places in some edge cases we still may need to make some of them public.
So yes please make one for Colors and UIColors also, and maybe deprecate the ones inside Config too so people stop using that and we can remove them later on.

Sources/Config.swift
```diff
@@ -1,4 +1,12 @@
 extension Warp {
+    public static var Token: TokenProvider {
+        return Config.tokenProvider
```
mohsen-schibsted, line 3: Maybe just copy and paste that code to here and just deprecate the ones inside Config so other people stop using those too 👍

mohsen-schibsted (approved):

Sources/Config.swift
dinsen: My plan was to get this into a new release.
Fix the NMP project and then deprecate/remove the old variables in a new Warp release after.
Just to ensure the project always can compile. But again. we have a summer release freeze in NMP, so maybe we should just make the cut? 🤔

mohsen-schibsted: I say just go for it 😅 should be fine. It is just a deprecation so only warnings even if someone is using it.

mohsen-schibsted: Looks great 🥳

PR #51, opened 2024-07-01
Change Spacing and Border to use let instead to avoid using rawValue on enum cases

dinsen:
I made a mistake when choosing to use enum cases. We need to get the value with `.rawValue` and that was not the intention.

Use changing all values to static let instead, just to keep it simple.

### DIFF
--- a/Sources/Border/Border.swift
+++ b/Sources/Border/Border.swift
@@ -1,40 +1,40 @@
 import CoreFoundation
 
 extension Warp {
-    public enum Border: CGFloat {
+    public enum Border {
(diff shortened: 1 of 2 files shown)


mohsen-schibsted (approved):

dinsen:
@mohsen-schibsted if you have the time doing the day, it will be fine 👌

mohsen-schibsted:
> @mohsen-schibsted if you have the time doing the day, it will be fine 👌

@dinsen release 0.0.12 is out there.

PR #54, opened 2024-07-02
Feature/expandable

pbodsk:
# Expandable Component

This looked like a simple component to implement to I went for it...of course I ended up spending more time than I thought 😅 

## How does it look? 🔍 
Here are a couple of videos showing the various designs of the `Expandable`. 

First with animations enabled:

https://github.com/warp-ds/warp-ios/assets/117636/e39aec05-0f56-4928-ba62-d5d556627abd

And then without animations:

https://github.com/warp-ds/warp-ios/assets/117636/b38c3e8a-aa7f-4784-9a45-062bb6fb8df0

## Reading Order 👀 
Here is a suggested reading order for the files in this PR

### `Expandable`
As always, a good first candidate is the component itself. It is not super complex and consists of two parts

- `alwaysVisibleView`: The top part that is always shown
- `expandableView`: The expanded/collapsed part sitting under the `alwaysVisibleView` part

To control expand/collapse we have a 

`@State private var isExpanded: Bool = false`

that is then toggled by a `.onTapGesture` on the `alwaysVisibleView`.

Note that we have to determine if the user wants this part to be animated or not:

```swift
.onTapGesture {
    if isAnimated {
        withAnimation(.easeInOut) {
            isExpanded.toggle()
        }
    } else {
         isExpanded.toggle()
    }
}
```

### `ExpandableStyle`

The `Expandable` can be one of three styles:

- `default`: Basic expandable component.
- `box`: Expandable component with a boxed layout. Round corners
- `boxBleed`:  Expandable component with a boxed layout. Square corners

So that calls for an `enum` which we can then also use to control `backgroundColor` and `cornerRadius`

### `ExpandableView`
Is a sandbox view demonstrating how to use the `Expandable` component

## Further Additions
I've taken the liberty of adding a `.navigationTitle` to all demo views...just for similarity 😄 

## That's it!
Yeah, not as complicated as some of the other components so I hope the review is more manageable as well 🤞 

Have fun 😅

### DIFF
--- a/Sources/Colors/ColorProvider.swift
+++ b/Sources/Colors/ColorProvider.swift
@@ -416,6 +416,7 @@ public struct ColorProvider {
     /// Expandable
     var expandableTitleText: Color { token.text }
     var expandableParagraphText: Color { token.text }
+    var expandableDefaultBackground: Color { token.background }
     var expandableBackground: Color { token.backgroundSubtle }
     var expandableBackgroundHover: Color { token.backgroundSubtleHover }
     var expandableIcon: Color { token.icon }
--- a/Sources/Colors/ColorProvider.swift
+++ b/Sources/Colors/ColorProvider.swift
@@ -416,7 +416,6 @@ public struct ColorProvider {
     /// Expandable
     var expandableTitleText: Color { token.text }
     var expandableParagraphText: Color { token.text }
-    var expandableDefaultBackground: Color { token.background }
     var expandableBackground: Color { token.backgroundSubtle }
     var expandableBackgroundHover: Color { token.backgroundSubtleHover }
     var expandableIcon: Color { token.icon }
--- a/Sources/Components/Expandable/Expandable.swift
+++ b/Sources/Components/Expandable/Expandable.swift
@@ -31,7 +31,6 @@ extension Warp {
          Starting in a collapsed state gives the user a high-level overview of the available information.
         */
         @State private var isExpanded: Bool = false
-        @State private var opacity: Double = 0.0
 
         private let fadeAnimationTime: Double = 0.1
         private let expandAnimationTime: Double = 0.2
@@ -97,28 +96,19 @@ extension Warp {
             .onTapGesture {
                 if isExpanded {
                     if isAnimated {
-                        withAnimation(.easeIn(duration: fadeAnimationTime)) {
-                            opacity = 0
-                        }
-
-                        withAnimation(.easeOut(duration: expandAnimationTime).delay(fadeAnimationTime)) {
+                        withAnimation(.easeInOut) {
                             isExpanded = false
                         }
                     } else {
-                        opacity = 0
                         isExpanded = false
                     }
                 } else {
                     if isAnimated {
-                        withAnimation(.easeIn(duration: expandAnimationTime)){
+                        withAnimation(.easeInOut) {
                             isExpanded = true
-                        }
-                        withAnimation(.easeOut(duration: fadeAnimationTime).delay(expandAnimationTime)) {
-                            opacity = 1
                         }
                     } else {
                         isExpanded = true
-                        opacity = 1
                     }
                 }
             }
@@ -137,7 +127,6 @@ extension Warp {
                 }
             }
             .fixedSize(horizontal: false, vertical: true)
-            .opacity(opacity)
         }
     }
 }
--- a/Sources/Components/Expandable/Expandable.swift
+++ b/Sources/Components/Expandable/Expandable.swift
@@ -1,30 +1,55 @@
 import SwiftUI
 
 extension Warp {
-    public struct Expandable: View{
-        @State var isExpanded: Bool = false
-        @State var opacity: Double = 0.0
+    /**
 
-        let fadeAnimationTime: Double = 0.1
-        let expandAnimationTime: Double = 0.2
+    Expandable allows users to interact with expandable/collapsible sections of content. Also known as “accordion”
+
+    **When to use**
+
+     Use the Expandable component when presenting content that can be optionally revealed to users.
+     Consider it for long content that may overwhelm users if displayed all at once.
+
+     **Usage**
+
+     Provide the `Warp.Expandable` with:
+
+     - A title.
+     - A subtitle.
+     - A `Warp.ExpandableStyle`, choose from:
+        - `default`: for a non bordered expandable.
+        - `box`: for a bordered expandable with rounded corners.
+        - `boxBleed`: for a bordered expandable with sharp corners.
+     - A boolean `isAnimated` to determine if the expand/collapse transition is animated or not. **Optional** _default value is `true` if none is specified_.
+     - A `ColorProvider`. **Optional:** _default is read from `Config.colorProvider` if none is specified_.
+     */
+    public struct Expandable: View {
+        /**
+        **Collapsed by default**
+
+         Expendables begins in the collapsed state with all content panels closed.
+         Starting in a collapsed state gives the user a high-level overview of the available information.
+        */
+        @State private var isExpanded: Bool = false
+        @State private var opacity: Double = 0.0
+
+        private let fadeAnimationTime: Double = 0.1
+        private let expandAnimationTime: Double = 0.2
 
         let title: String
         let subtitle: String
         let style: Warp.ExpandableStyle
         let isAnimated: Bool
 
-        /// Object responsible for providing colors in different environments and variants.
-        var colorProvider: ColorProvider = Config.colorProvider
+        let colorProvider: ColorProvider
 
         public init(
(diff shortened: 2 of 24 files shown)


Sources/Resources/Assets.xcassets/Expandable/icon-expandable-chevron.imageset/Contents.json
```diff
@@ -0,0 +1,12 @@
+{
+  "images" : [
+    {
+      "filename" : "icon-expandable-chevron.svg",
+      "idiom" : "universal"
```
dinsen, line 5: Should this be enabled?
Skærmbillede 2024-07-02 kl  11 28 50

Sources/Components/Expandable/Expandable.swift
```diff
@@ -0,0 +1,142 @@
...
+                    .foregroundColor(colorProvider.expandableIcon)
+                    .frame(width: 16, height: 16)
+                    .rotationEffect(.degrees(isExpanded ? 180 : 0))
+
+                if style == .default {
+                    Spacer()
+                }
+            }
+            .contentShape(Rectangle())
+            .onTapGesture {
```
dinsen, line 93: `.onTapGesture` vs `Button` I think I've seen some tweets about getting some free accessibility stuff if you use a button instead of .onTapGesture 🤔 

Do you have any pros/cons ?

dinsen (approved):
This looks really good 🚀

Sources/Colors/ColorProvider.swift
```diff
@@ -416,6 +416,7 @@ public struct ColorProvider {
     /// Expandable
     var expandableTitleText: Color { token.text }
     var expandableParagraphText: Color { token.text }
+    var expandableDefaultBackground: Color { token.background }
```
mohsen-schibsted, line 419: Adding a color to ColorProvider without it being part of WARP colors or tokens will remove it next time we sync with Figma/Taxonomy sheet. So you should avoid doing that.

And I beleive the default expandable background color is `.clear` not token.background which will be `.white` in light mode and `.gray900` in dark mode.

So my suggestion is remove this colors, and use .clear in your `ExpandableStyle`

Sources/Components/Expandable/Expandable.swift
```diff
@@ -0,0 +1,142 @@
...
+        /**
+        **Collapsed by default**
+
+         Expendables begins in the collapsed state with all content panels closed.
+         Starting in a collapsed state gives the user a high-level overview of the available information.
+        */
+        @State private var isExpanded: Bool = false
+
+        let title: String
+        let subtitle: String
```
milad-schibsted, line 36: On other platforms, it's possible to pass a view to the Expendable component. Should we do the same, or is a string of text enough ?

Sources/Components/Expandable/Expandable.swift
```diff
@@ -0,0 +1,142 @@
...
+                    title,
+                    style: .bodyStrong,
+                    color: colorProvider.expandableTitleText
+                )
+
+                if style != .default {
+                    Spacer()
+                }
+
+                Image("icon-expandable-chevron", bundle: .module)
```
milad-schibsted, line 82: It's possible to use the SFSymbol 'chevron.down' for this.

Sources/Colors/ColorProvider.swift
pbodsk: Ah, I was not aware 😄 I've changed it to `.clear` now 👌

Sources/Components/Expandable/Expandable.swift
pbodsk: I can make it so we take a `ViewBuilder` as parameter...I think I'll make both available then, so that you can pass both a `String` and an entire View. Let me give that a go.

Sources/Resources/Assets.xcassets/Expandable/icon-expandable-chevron.imageset/Contents.json
pbodsk: Good question. It is not at the moment and doesn't seem to be in any of the icons in the Assets folder.

From what I read here:

https://forums.developer.apple.com/forums/thread/694526

> When i tick the 'Preserve Vector Data' box, the asset is scaled at runtime using the vector data, allowing for smooth scaling and crisp images at any scale.

So that kinda indicate that it should be checked...but then we should probably do that on all images.

Have you had any discussions previously related to this @mohsen-schibsted?

Sources/Components/Expandable/Expandable.swift
pbodsk: I started out with having this being a button but there was a problem with the animations looking wonky if I did so. Let me try again because I've made _a lot_ of changes since then, so maybe some of them has done something good for me.

Sources/Resources/Assets.xcassets/Expandable/icon-expandable-chevron.imageset/Contents.json
mohsen-schibsted: No we have not started working on any icons. Our designers are working on our icon v2.0 and when that is ready we will start.

Sources/Components/Expandable/Expandable.swift
pbodsk: Ah yes...there was the problem with the button highlighting on taps even with `.buttonStyle(.plain)` set.

Fixed that with a new private `NonHighlightedButtonStyle`

Sources/Resources/Assets.xcassets/Expandable/icon-expandable-chevron.imageset/Contents.json
pbodsk: Right, I'll wait till the icons have settled then 👍

Sources/Components/Expandable/Expandable.swift
pbodsk: You can see how it looks in [the most recent commit](https://github.com/warp-ds/warp-ios/pull/54/commits/dca6d67569fa7aa6308c821a002f3ae19a423f3e)

Sources/Components/Expandable/Expandable.swift
pbodsk: But do we want that? Don't we want to use our own?

Sources/Components/Expandable/ExpandableStyle.swift
```diff
@@ -0,0 +1,30 @@
...
+                .clear
+            case .box, .boxBleed:
+                colorProvider.expandableBackground
+            }
+        }
+        
+        var cornerRadius: Double {
+            switch self {
+            case .default, .boxBleed: 0
+            case .box: 8
```
mohsen-schibsted, line 26: Should we use `Warp.Border.borderRadius100`?

Sources/Components/Expandable/Expandable.swift
```diff
@@ -0,0 +1,201 @@
...
+       - A `subtitle` String
+       - An `expandableView` `View`
+     - A boolean `isAnimated` to determine if the expand/collapse transition is animated or not. **Optional** _default value is `true` if none is specified_.
+     - A `ColorProvider`. **Optional:** _default is read from `Config.colorProvider` if none is specified_.
+     */
+    public struct Expandable<Content: View>: View {
+        /**
+         **Collapsed by default**
+
+         Expendables begins in the collapsed state with all content panels closed.
```
mohsen-schibsted, line 32: It seems on web component can start from expanded state also, so maybe we should follow suit and add that as a param.

Screenshot 2024-07-03 at 10 21 25

Sources/Components/Expandable/Expandable.swift
mohsen-schibsted: We have had some discussions around this and what's the best practice. If we use SFSymbol then we get some accessibility features for free and for example the icon will be dynamic and get bigger sizes. But in some places we may not want that.
Since we have not started defining icons we don't have a preference here. Maybe we will later on.

For now I would say if it makes sense for the icon to stay the same size use custom icon, if it make sense for it to be dynamic use SFSymbol.

Sources/Components/Expandable/Expandable.swift
mohsen-schibsted: Looks great 👍

Sources/Components/Expandable/ExpandableStyle.swift
pbodsk: Yeah why not...now that we have it 😅 

I'll update

Sources/Components/Expandable/Expandable.swift
pbodsk: Ha! I had initially exposed the parameter so you could start expanded 😅 

No problem, I'll do so again 👍

mohsen-schibsted (approved):

PR #64, opened 2024-08-22
Create run-tests.yml

mohsen-schibsted:
Add github action


.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,24 @@
+# This workflow will build a Swift project
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-swift
+
+name: Xcode - Build and Test
+
+on:
+  push:
+    branches: [ "main" ]
```
vishkaischibsted, line 8: Just for clarification,
Do you really need this run on every push?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,24 @@
...
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  build:
+
+    runs-on: macos-14
```
vishkaischibsted, line 15: Personally I would like to recommend using the latest macos, but combined with Xcode specification.

luk-m-s: Done

PR #65, opened 2024-08-28
Radio component

mohsen-schibsted:
https://www.figma.com/design/oHBCzDdJxHQ6fmFLYWUltf/WARP---Components-2.0?node-id=381-40974&t=5ky9C3LxbmD5qNeu-0


Sources/Components/Radio/Radio.swift
```diff
@@ -0,0 +1,136 @@
...
+    ///
+    /// `Radio` is a customizable component that allows users
+    /// to create a list of radio buttons where only one can be selected at any given time.
+    ///
+    /// - Parameters:
+    ///   - selectedOption: A binding to the currently selected option.
+    ///   - options: An array of options that conform to `Identifiable` and `Hashable`.
+    ///   - label: A closure that provides a label for each option.
+    ///   - state: The state of the radio button group (default, error, disabled).
+    public struct Radio<Option: Identifiable & Hashable>: View {
```
pbodsk, line 14:
```suggestion
    public struct RadioButtonGroup<Option: Identifiable & Hashable>: View {
```

It seems you had your mind set on RadioButtonGroup previously in a comment:

> A single radio button view used within `RadioButtonGroup`.

And I like that better actually. That or at least something that conveys that this component is not a single radio button...it is a list of radio buttons 😄

Sources/Components/Radio/Radio.swift
```diff
@@ -0,0 +1,136 @@
...
+        /// A binding to the currently selected option.
+        @Binding var selectedOption: Option
+        /// An array of options that conform to `Identifiable` and `Hashable`.
+        var options: [Option]
+        /// A closure that provides a label for each option.
+        var label: (Option) -> String
+        /// The state of the radio button group (default, error, disabled).
+        var state: RadioButtonState
```
pbodsk, line 22:
```suggestion
        var radioButtonState: RadioButtonState
```
🤔 is there a chance that people will confuse `state` for something that is expected to be a `@State` variable?

Sources/Components/Radio/Radio.swift
```diff
@@ -0,0 +1,136 @@
...
+        
+        public var body: some View {
+            VStack(alignment: .leading, spacing: Spacing.spacing200) {
+                ForEach(options) { option in
+                    RadioButton(isSelected: selectedOption == option,
+                                label: label(option),
+                                state: state) {
+                        if state != .disabled {
+                            selectedOption = option
+                        }
```
pbodsk, line 51: You seem to have the same check inside the `RadioButton` itself:

```swift
.onTapGesture {
    if state != .disabled {
        action()
    }
}
```

Is that intentional?

Sources/Components/Radio/Radio.swift
```diff
@@ -0,0 +1,136 @@
...
+            case (.default, false):
+                return colorProvider.radioBorder
+            case (.error, true):
+                return colorProvider.radioNegativeBorder
+            case (.error, false):
+                return colorProvider.radioNegativeBorder
+            case (.disabled, true):
+                return colorProvider.radioBorderDisabled
+            case (.disabled, false):
+                return colorProvider.radioBorderDisabled
```
pbodsk, line 113:
```suggestion
            case (.error, _):
                return colorProvider.radioNegativeBorder
            case (.disabled, _):
                return colorProvider.radioBorderDisabled
```

Just...in case you don't care about it being selected or not in these scenarios 🤷

Sources/Components/Radio/Radio.swift
```diff
@@ -0,0 +1,136 @@
...
+            }
+            .onTapGesture {
+                if state != .disabled {
+                    action()
+                }
+            }
+        }
+        
+        private var borderColor: Color {
+            switch (state, isSelected) {
```
pbodsk, line 101: Uuuuh...clever...I like it ❤️ I always forget about the "multi variable" `switch`

Sources/Components/Radio/Radio.swift
```diff
@@ -0,0 +1,136 @@
...
+            }
+        }
+        
+        private var fillColor: Color {
+            if state == .disabled {
+                colorProvider.radioBackgroundDisabled
+            } else {
+                .clear
+            }
+        }
```
pbodsk, line 123:
```suggestion
        private var fillColor: Color {
            state == .disabled ? colorProvider.radioBackgroundDisabled : .clear
        }
```

You know...if you want to boil it down 😄

Sources/Components/Radio/Radio.swift
```diff
@@ -0,0 +1,136 @@
...
+            switch state {
+            case .default:
+                return colorProvider.token.text
+            case .error:
+                return colorProvider.token.text
+            case .disabled:
+                return colorProvider.token.textDisabled
+            }
+        }
+    }
```
pbodsk, line 135:
```suggestion
        private var textColor: Color {
            switch state {
            case .default, .error:
                return colorProvider.token.text
            case .disabled:
                return colorProvider.token.textDisabled
            }
        }
    }
```

I can't determine if I like the explicit-y of showing all states or combining the cases with the same output into one.

You decide 😄

dinsen (approved):
Looks good @mohsen-schibsted  👌

pbodsk (approved):
Looking very nice :clap: I've left some thoughts and suggestions here and there..just the usual nitpicking you know 😅 

## An Upcoming Feature? 
I thought about...maybe not for version 1.0 but at some point someone will come with a design like this:

Screenshot 2024-08-29 at 10 36 55

Or something even more crazy (think multiline text with different fonts) and expect to use "our" radio buttons for that. 

And they can't at the moment since we only have a text representation.

So...a potential future feature would be that you can provide your own `View`(similar to how the `label` parameter is used on a `Button` for instance).

A first step could be to provide something like this:

Screenshot 2024-08-29 at 10 36 55

Where:

- the green box is provided by Warp.Radio
- the red box is where the developer can roam free and make what they want

So...what we have now, and then with a `@ViewBuilder label: () -> Label` parameter.

Hope that makes sense...and as said, this could be a 1.1. feature.

Again...good work :clap:

mohsen-schibsted:
> Looking very nice 👏 I've left some thoughts and suggestions here and there..just the usual nitpicking you know 😅
> 
> ## An Upcoming Feature?
> I thought about...maybe not for version 1.0 but at some point someone will come with a design like this:
> 
> Screenshot 2024-08-29 at 10 36 55
> 
> Or something even more crazy (think multiline text with different fonts) and expect to use "our" radio buttons for that.
> 
> And they can't at the moment since we only have a text representation.
> 
> So...a potential future feature would be that you can provide your own `View`(similar to how the `label` parameter is used on a `Button` for instance).
> 
> A first step could be to provide something like this:
> 
> Screenshot 2024-08-29 at 10 36 55
> 
> Where:
> 
> * the green box is provided by Warp.Radio
> * the red box is where the developer can roam free and make what they want
> 
> So...what we have now, and then with a `@ViewBuilder label: () -> Label` parameter.
> 
> Hope that makes sense...and as said, this could be a 1.1. feature.
> 
> Again...good work 👏

We talked about the border, That will be something we will create later, won't be a `RagioGroup` with border. But maybe a new component created by combining `CardComponent` and `RadioComponent`

Sources/Components/Radio/Radio.swift
mohsen-schibsted: Good point, I am gonna remove it from the group

Sources/Components/Radio/Radio.swift
mohsen-schibsted: not that it is not `.clear` any more maybe it will be long and more confusing

Sources/Components/Radio/Radio.swift
mohsen-schibsted: yeah I like the simplicity of your suggestion. When you are writing it you change things one by one so you don't see it later that they are the same color

PR #71, opened 2024-09-20
Add support for icon

mohsen-schibsted:
Add icons from [here](https://www.figma.com/design/yEx16ew6S0Xgd579dN4hsM/WARP---Icons?node-id=0-1&node-type=canvas&t=xlrURigNZir4xZ68-0) and IconView component with 3 sizes and a custom size


Sources/Components/IconView/IconSize.swift
```diff
@@ -0,0 +1,46 @@
+import SwiftUI
+
+extension Warp {
+    /// Icon sizes that can be used in `IconView`.
+    ///
+    /// `IconSize` defines three standard sizes: `small` (16x16), `default` (24x24), `large` (32x32),
+    /// and a `custom` size that can be specified by the user.
+    public enum IconSize: CaseIterable, Hashable {
+        public static var allCases = [Warp.IconSize.small, .default, .large]
```
vishkaischibsted, line 9: This property cam be omitted, since it will be generated by compiler.

Sources/Components/IconView/IconSize.swift
```diff
@@ -0,0 +1,46 @@
...
+            case .small:
+                return "Small"
+            case .default:
+                return "Default"
+            case .large:
+                return "Large"
+            case .custom(_):
+                return "Custom" // Display custom size in points
+            }
+        }
```
vishkaischibsted, line 44: Should this property exists on the IconSize itself?

Sources/Icon/Icon.swift
```diff
@@ -0,0 +1,275 @@
...
+                .renderingMode(.template) // Ensure template rendering mode for vector images
+                .resizable()
+        }
+        
+        /// Returns a `UIImage` for the corresponding icon, loaded from the asset catalog.
+        ///
+        /// - Returns: A `UIImage` object corresponding to the icon.
+        public var uiImage: UIImage {
+            UIImage(named: assetName, in: .module, compatibleWith: nil) ?? UIImage() // Load the image from the asset catalog
+        }
```
vishkaischibsted, line 248:
```suggestion
        public var uiImage: UIImage? {
            UIImage(named: assetName, in: .module, compatibleWith: nil)
        }
```

Sources/Icon/Icon.swift
```diff
@@ -0,0 +1,275 @@
...
+            let originalImage = uiImage
+            
+            let targetSize = CGSize(width: width, height: height)
+            UIGraphicsBeginImageContextWithOptions(targetSize, false, 0.0)
+            originalImage.draw(in: CGRect(origin: .zero, size: targetSize))
+            let resizedImage = UIGraphicsGetImageFromCurrentImageContext()
+            UIGraphicsEndImageContext()
+            
+            return resizedImage ?? UIImage()
+        }
```
vishkaischibsted, line 266:
```suggestion
        public func resizeImage(to targetSize: CGSize) -> UIImage? {
            guard let originalImage = uiImage else {
                return
            }
            
            UIGraphicsBeginImageContextWithOptions(targetSize, false, 0.0)
            originalImage.draw(in: CGRect(origin: .zero, size: targetSize))
            let resizedImage = UIGraphicsGetImageFromCurrentImageContext()
            UIGraphicsEndImageContext()
            
            return resizedImage
        }
```

Sources/Components/IconView/IconSize.swift
mohsen-schibsted: allCases? No it won't. It has a custom one also :D which has a associated value.

Sources/Icon/Icon.swift
mohsen-schibsted: I don't want it to be optional. We control the image so we make sure that it will never go to the default case

Sources/Components/IconView/IconSize.swift
vishkaischibsted: Then it's not really `CaseIterable`, right? =D

mohsen-schibsted: That's on Apple not us 😅 That custom case is not part of Warp really, it is just there to make sure that users can override it if needed.

Maybe in this case it makes sense to instead of having it confirm to `CaseIterable` just have it on the demo app, since `allCases` is only for demoing purposes.

Sources/Icon/Icon.swift
mohsen-schibsted: One idea is to use a guard and return UIImage in there, so we can later send an error when we have our Crashlytics setup too 🤔

Sources/Components/IconView/IconSize.swift
mohsen-schibsted: What do you mean? it is on the IconSize itself

Sources/Components/IconView/IconSize.swift
```diff
@@ -0,0 +1,46 @@
...
+            case .small:
+                return 16
+            case .default:
+                return 24
+            case .large:
+                return 32
+            case .custom(let size):
+                return size
+            }
+        }
```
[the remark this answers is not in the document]
mohsen-schibsted: If we change the name then it will be `.frame(width: size.size, height: size.size)` maybe a little hard to read

Sources/Components/IconView/IconSize.swift
vishkaischibsted: `IconSize`, the type itself!

PR #82, opened 2024-10-14
Create label component

Ali-NaranjiSani:
Add support for [Label](https://www.figma.com/design/oHBCzDdJxHQ6fmFLYWUltf/WARP---Components?node-id=401-35803&node-type=canvas&t=78BUGrlfdL8KjRG1-0).

### DIFF
--- a/Sources/Components/Label/Label.swift
+++ b/Sources/Components/Label/Label.swift
@@ -69,7 +69,7 @@ extension Warp {
         // MARK: - Body
 
         public var body: some View {
-            HStack {
+            HStack(spacing: 4) {
                 titleView
                 optionalLabelView
                 toolTipView
--- a/Sources/Components/Label/Label.swift
+++ b/Sources/Components/Label/Label.swift
@@ -2,6 +2,26 @@ import Foundation
 import SwiftUI
 
 extension Warp {
+    /// A customizable label component used to display a title, optional additional information, and an info tooltip.
+    ///
+    /// The `Warp.Label` component displays a main title with optional additional information. It can also show an info icon which, when tapped, presents custom tooltip content provided by the user.
+    ///
+    /// **Usage:**
+    ///
+    /// ```swift
+    /// Warp.Label(
+    ///     title: "Label Title",
+    ///     additionalInformation: "Optional Info",
+    ///     showToolTipImage: true,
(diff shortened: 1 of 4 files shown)


Sources/Components/Label/Label.swift
```diff
@@ -0,0 +1,147 @@
+import Foundation
```
mohsen-schibsted, line 1: We don't need to import Foundation, it is part of SwiftUI

Sources/Components/Label/Label.swift
```diff
@@ -0,0 +1,147 @@
...
+    ///
+    /// The `Warp.Label` component displays a main title with optional additional information. It can also show an info icon which, when tapped, presents custom tooltip content provided by the user.
+    ///
+    /// **Usage:**
+    ///
+    /// ```swift
+    /// Warp.Label(
+    ///     title: "Label Title",
+    ///     additionalInformation: "Optional Info",
+    ///     showToolTipImage: true,
```
mohsen-schibsted, line 15: Everywhere else we are using Tootip, but here it is `showToolTipImage` maybe it should be `showTooltipImage`

Sources/Components/Label/Label.swift
```diff
@@ -0,0 +1,146 @@
...
+    ///     tooltipContent: AnyView(Warp.Tooltip(title: "Tooltip text"))
+    /// )
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - title: The main title text to display.
+    ///   - additionalInformation: Optional text to display after the title.
+    ///   - showToolTipImage: Indicates whether to show the tooltip icon.
```
mohsen-schibsted, line 22: here also `showToolTipImage` should be `showTooltipImage`, and the other occurances

Sources/Components/Label/Label.swift
```diff
@@ -0,0 +1,146 @@
...
+            self.title = title
+            self.additionalInformation = additionalInformation
+            self.showToolTipImage = showToolTipImage
+            self.tooltipContent = tooltipContent
+        }
+
+        // MARK: - Body
+
+        public var body: some View {
+            HStack(spacing: 4) {
```
mohsen-schibsted, line 71: Maybe use `Warp.spacing50`

mohsen-schibsted (approved):

PR #84, opened 2024-10-16
Add color provider

mohsen-schibsted:
We are getting the ColorProvider from [Style Dictionary](https://github.com/warp-ds/tokens) too.

And some cleanup

### DIFF
--- a/Sources/Colors/BrandsSampleView/BlocketColorsView.swift
+++ b/Sources/Colors/BrandsSampleView/BlocketColorsView.swift
@@ -2,7 +2,7 @@ import SwiftUI
 
 public struct BlocketColorsView: View {
     public init() {}
-
+    
     public var body: some View {
         ScrollView(showsIndicators: false) {
             VStack (alignment: .leading) {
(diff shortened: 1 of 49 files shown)


Sources/Colors/BrandsSampleView/BlocketColorsView.swift
```diff
@@ -2,7 +2,7 @@ import SwiftUI
 
 public struct BlocketColorsView: View {
     public init() {}
-
+
```
Ali-NaranjiSani, line 5: There are some white spaces in the following lines that seem unnecessary. These changes aren't needed, right?

Sources/Components/Alert/Alert.swift
```diff
@@ -212,57 +215,57 @@ extension Warp {
 
 private struct ButtonsView: View, Hashable {
     let primaryButtonProvider: Warp.Alert.ButtonConstructor?
-
+
```
Ali-NaranjiSani, line 218: Same white spaces here.

Ali-NaranjiSani (approved):
Nicely done Mohsen.
The documentation added is excellent!

Sources/Colors/BrandsSampleView/BlocketColorsView.swift
mohsen-schibsted: I used `cmd+A` to select all and then `ctrl+I` to reindent to whatever `Xcode` decided is the right indentation 😅
So probably was wrong before

PR #86, opened 2024-10-18
Create TextArea component

Ali-NaranjiSani:
- Add support for [TextArea](https://www.figma.com/design/oHBCzDdJxHQ6fmFLYWUltf/WARP---Components?node-id=1308-38523&node-type=canvas&t=izfmza926rHBhwGR-0).

- some tiny enhancements on HelpTextView


Sources/Components/TextArea/TextAreaStyle.swift
```diff
@@ -0,0 +1,9 @@
+extension Warp {
+    /// Represents the different styles a `TextArea` can have.
+    public enum TextAreaStyle: String, CaseIterable {
+        case `default`
+        case disabled
+        case error
+        case readOnly
+    }
```
Ali-NaranjiSani, line 8: Also, there is one more style, 'Active,' which shouldn't be accessible to the user, so I didn't define it here. In the Active style, we highlight the border, which is handled internally by the component.

mohsen-schibsted (approved):

PR #94, opened 2024-11-04
Create switch component

Ali-NaranjiSani:
- Add support for [Switch](https://www.figma.com/design/oHBCzDdJxHQ6fmFLYWUltf/WARP---Components?node-id=915-37468&node-type=canvas&t=reKP2L0yr5wDelzK-0).


SampleApp/Components/SwitchView.swift
```diff
@@ -0,0 +1,53 @@
...
+            
+            // Controls to modify Switch's state
+            GroupBox(content: {
+                VStack(alignment: .leading) {
+                    // Toggle Switch State
+                    HStack {
+                        Text("Toggle Switch State")
+                        Spacer()
+                        
+                        Warp.Button.create(
```
mohsen-schibsted, line 26: Why is it a button, can't we use the Warp.Switch here?

mohsen-schibsted: Warp.Switch(isOn: $isDisabled)

Or maybe a segment

Picker("Enable or Disable", selection: $isDisabled) {
    Text("Enable").tag(false)
    Text("Disable").tag(true)
}
.pickerStyle(SegmentedPickerStyle())

mohsen-schibsted (approved):

PR #95, opened 2024-11-07
Add public SwiftUI.Image accessor to Warp.Icon

Breiby:
# Why?

Sometimes one might want to create a `View` that takes an `Image` argument. In order to use Warp icons with such views, we need an `Image` accessor as well.

# What?

Added an accessor to Warp icons so that one can get the corresponding SwiftUI image like `Warp.Icon.someIcon.image`. Have purposefully not touched `BrandIcon` or `TaxonomyIcon` as I don't foresee they have the same needs.

Downside of this is that we lose the accessibility label, but as far as I can tell it's just reading the asset name raw value anyway so it should probably be handled by the user of the icon anyway?

### DIFF
--- a/Sources/Icon/Icon.swift
+++ b/Sources/Icon/Icon.swift
@@ -244,10 +244,17 @@ extension Warp {
         ///
         /// - Returns: A resizable `SwiftUI.Image` instance representing the icon, with a template rendering mode.
         public var body: some View {
+            image
+                .accessibilityLabel(localization)
+        }
+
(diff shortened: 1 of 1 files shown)


mohsen-schibsted:
Thank you @Breiby for the PR. For now yes the AccessibilityLabel reads the raw value, but we have another branch that we are working on to get the translations from Crowdin and map the accessibility to those which hopefully we can merge later on. I will take care of the accessibilityLabel when that one is ready.

mohsen-schibsted (approved):

PR #97, opened 2024-11-08
Fix icon raw values

mohsen-schibsted:
Add some custom rawValues to icons so we are in sync with web and Android. Add the rendering mode inside the assets also.


Sources/Helpers/Image+.swift
```diff
@@ -0,0 +1,9 @@
+import SwiftUI
+
+extension Image {
+    /// Loads an image from the Warp bundle.
+    /// - Parameter named: The rawValue of the icon from `Warp Assets`.
+    public static func warpIcon(named name: String) -> Image {
```
adriansergheev, line 6: isn't it a little better to have this namespaced to  `Warp`?
ex: `Warp.icon(name: String)`
it's ok it this way as well, but it does seem that all other endpoins Warp provides are under Warp namespace and not extensions to native types

adriansergheev (approved):

Sources/Helpers/Image+.swift
mohsen-schibsted: Poor Xcode does not know what to use without more context 😅

Screenshot 2024-11-08 at 15 38 15

So I will leave it like this since we have other public extensions like cardBackground and roundedCorners and the like.

PR #117, opened 2024-12-20
Add support for foregroundStyle to Text component

mohsen-schibsted:
Add support for foregroundStyle to Text component

This way we can add gradients

Simulator Screenshot - iPhone 16 Pro - 2024-12-20 at 12 49 16


nssina (approved):
🔥

PR #118, opened 2025-01-03
Improvement: set `TokenProvider` and `UITokenProvider` when the `Theme` is updated

robinsalehjan:
# Why?
- Previously, changing the `Theme` property between test runs did not reliably update associated colors and tokens, leading to incorrect themes being applied when snapshot tests ran in parallel. 

# What?

- Change the properties: `Token`, `UIToken`, `Color`, `UIColor`, `DatavizToken` and `DatavizUIToken` from `stored` properties to `computed` properties. These properties will be re-computed whenever the `Theme` property is changed rather than computed and stored once.

- Validate that the font has not been registered before registering it.

- Make `Theme` property conform to `CustomStringConvertible` and `CaseIterable` in order to use it as a parameter that we can pass to the parameterised tests provided from the `Swift Testing` framework.

### DIFF
--- a/Sources/Typography/Typography+Register.swift
+++ b/Sources/Typography/Typography+Register.swift
@@ -30,13 +30,22 @@ extension Warp.Typography {
             throw Warp.FontRegistrationError.unableToFindFont(name: fontName)
         }
         
-        // Attempt to register the font with Core Text
-        var error: Unmanaged<CFError>?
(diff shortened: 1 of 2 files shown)


robinsalehjan:
> Looks great @robinsalehjan 👍 have you tested this on your tests and achieved what you intended to?

Yes here a recording with the parameterised tests running once for each brand👍 

https://github.com/user-attachments/assets/9703a506-e6e1-41b7-bb48-24fb7752cc29

PR #130, opened 2025-03-22
Add scrollable parameter to Tabs component

Breiby:
# Why? 

We want to be able to create screens with just two tabs with the tabs taking up the full width of the screen regardless of text length.

# What?

Add a parameter `scrollable` to the `Tabs` component which if set to `false` neglects the `ScrollView` and makes the tabs take up the available horizontal space instead. 

### Note
This change does not handle long text lengths or changes to text size due to dynamic type. The existing view doesn't really handle dynamic type well either, so I'm not sure I'd consider this a flaw of this PR in particular.

# Screenshot

Screenshot 2025-03-22 at 14 43 59

### DIFF
--- a/SampleApp/Components/TabsView.swift
+++ b/SampleApp/Components/TabsView.swift
@@ -3,6 +3,7 @@ import Warp
 
 struct TabsView: View {
     @State private var hasIcons = true
+    @State private var isScrollable = true
     @State private var selectedIndex: Int = 0
     private let tabItems: [Warp.TabItem] = [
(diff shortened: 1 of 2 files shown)


pbodsk (commented):
Looks good to me and funny enough, we just had the _exact_ same use case so...glad to see this implemented 👏

mohsen-schibsted (approved):

mohsen-schibsted:
Could you please add this to the `SampleApp` also? `TabsView`

PR #136, opened 2025-04-24
Add theme preview modifier

Breiby:
# Why?

Xcode previews in the SMP project look wrong because Warp theme is not set

# What?

Add a preview modifier that lets us set the theme like:
```swift
@available(iOS 18.0, *)
#Preview(traits: .themeFinn) {
    MyWarpReliantView()
}
```


bstien (approved):
Nice!🙌🥇

adriansergheev (approved):

PR #137, opened 2025-05-01
Add content shape to tab item

Breiby:
# Why?

If the tab item is wider than its text, the area outside of the text does not respond to taps

# What?

Give the tab item a content shape. Also remove a wrapping VStack which seemed superfluous.


pbodsk (approved):
Looking good :clap:

I often forget about `.contentShape` when building buttons. It seems like a bit of an oversight that we have to add this ourselves to be honest but...here we are. Nice catch :shipit:

PR #141, opened 2025-05-13
Make CheckBox title optional

AtishPapa:
# Why, What, How

There's a new component in TJT and I wanted to use the check box from Warp to conform to the unified design principles. At the moment it always requires a title and setting and empty `String` would not result in an optimal layout.

In this PR I made the title optional and set `nil` as a default value. This won't break the current implementations.

| Before | After |
| --- | --- |
| Screenshot 2025-05-12 at 16 55 54 | Screenshot 2025-05-12 at 16 56 14 |

As you can see on the screenshots the view boundaries are looking cleaner.

# A small detail

Having shown the disabled state I noticed that the checkmark color is barely visible so I set a more noticeable one.

| Before | After |
| --- | --- |
| Screenshot 2025-05-13 at 10 28 38 | Screenshot 2025-05-13 at 10 29 10 |

### DIFF
--- a/Sources/Components/Checkbox/Checkbox.swift
+++ b/Sources/Components/Checkbox/Checkbox.swift
@@ -35,7 +35,7 @@ extension Warp {
         ///   - extraContent: An optional view that will be displayed beside the label.
         ///   - action: A closure that is executed when the checkbox is tapped.
         public init(isSelected: Binding<Bool>,
-                    label: String? = nil,
+                    label: String,
                     style: CheckboxStyle = .default,
                     extraContent: AnyView? = nil,
                     action: @escaping () -> Void) {
@@ -45,7 +45,21 @@ extension Warp {
             self.extraContent = extraContent
             self.action = action
         }
-        
+
+        /// Initializes a new `Checkbox` without title and extra content.
+        ///
+        /// - Parameters:
+        ///   - isSelected: A `Binding` value indicating whether the checkbox button is selected.
+        ///   - style: The style the checkbox can have (default, error, disabled).
+        public init(isSelected: Binding<Bool>,
+                    style: CheckboxStyle = .default) {
+            self._isSelected = isSelected
+            self.label = nil
+            self.style = style
+            self.extraContent = nil
+            self.action = { }
+        }
+
         public var body: some View {
             HStack(alignment: .top, spacing: Spacing.spacing100) {
                 Image(iconName, bundle: .module)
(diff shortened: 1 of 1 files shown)


mohsen-schibsted (approved):

Sources/Components/Checkbox/Checkbox.swift
```diff
@@ -45,7 +45,21 @@ extension Warp {
...
             self.action = action
         }
-        
+
+        /// Initializes a new `Checkbox` without title and extra content.
+        ///
+        /// - Parameters:
+        ///   - isSelected: A `Binding` value indicating whether the checkbox button is selected.
+        ///   - style: The style the checkbox can have (default, error, disabled).
+        public init(isSelected: Binding<Bool>,
```
mohsen-schibsted, line 54: Can we achieve the same thing by just adding a default value on line 41?
`action: @escaping () -> Void = {}) {`

AtishPapa: Of course we can, but I found this solution a bit more descriptive. Do you prefer making the `title` optional with default `nil` and the action with a default empty closure? 🙂

mohsen-schibsted: I would go with one initializer, easier to document with the structure we have in our doc.

AtishPapa: OK, I've made the changes according to that.

mohsen-schibsted: Looks great thanks 👍
Feel free to update docs also if you know how to do it.
https://github.com/warp-ds/docs/blob/main/docs/components/checkbox/ios.md
https://github.com/warp-ds/docs/blob/1b36bb81ba3e23ec6c0709e480cc01d086931380/docs/api-table.js#L3307
Otherwise I can do it.

How about `Radio` component? Should we do the same?

AtishPapa: Uhh, I don't really know that documentation format. 😬 Could you do the changes?

Regarding the radio it's not urgent at this time, but I'll probably do it soon.

mohsen-schibsted: Amazing, yes I will update the docs

mohsen-schibsted: are you releasing this change also?

AtishPapa: Do we have a script or automation for that? 🙂

mohsen-schibsted: Nope, fully fully manual :D

PR #142, opened 2025-05-16
Fonts: Add italic FINN fonts

bstien:
# Why?
We have need for italic fonts in FINN, since the Job verticals allows for italic fonts in both their import and their editor.

We discussed using a regular system font as substitution for italic fonts, but since we already have `FINNType` in italic why not use that?

# What?
- Added italic font files for light and medium weight.
- Added font descriptions to `Warp.Font`.


Sources/Typography/Font.swift
```diff
@@ -99,7 +105,7 @@ extension Warp {
         static var fontForTheme: [Font] {
             switch Theme {
             case .finn:
-                return [.finnLight, .finnMedium]
+                return [.finnLight, .finnLightItalic, .finnMedium, .finnMediumItalic]
```
Breiby, line 108: Curious that FINN is the only brand without a **bold** font

Breiby (approved):

bstien:
The initial PR had italic fonts for both light and medium weights. iOS isn't able to map which font/font file belongs together, so we encountered an issue where a view using `light` font would use `medium-italic` as soon as we asked for the corresponding font with italic traits. A random guess is that the fonts might not have enough metadata for iOS to properly figure out which ones belong together.

And we have very limited options when it comes to registering custom fonts on the system, especially when it comes to fonts/traits-mapping, so to get this out and fix the issue we have we're dropping `medium-italic` for now.

PR #164, opened 2025-09-18
Fix: Double padding in `TextField` component (WARP-1097)

luk-m-s:
## Why? ##

Previous implementation doesn't use HStack spacing but add horizontal padding to each of `TextField` component. This cause **double padding** between text input and border (or leading element).

## Snapshots ##
Before: 
image
After:
image


ritacoimbra (commented):
From a visual perspective, the before and after snapshots look great! 👌

PR #199, opened 2026-03-05
Navigation bar Liquid Glass style modifiers

luk-m-s:
### UIKit

Set up global appearance for navigation bar
```swift
UINavigationBar.warpLiquidGlassStyle()
```
Set up appearance for concrete instance
```swift
UINavigationBar().warpLiquidGlassStyle()
// alternatively
UIViewController().warpLiquidGlassStyle()
```

Set up style appearance for UIBarButtonItem
```swift
UIBarButtonItem(
     image: Warp.Icon.foo.uiImage,
     style: .plain,
     target: self,
      action: nil
)
    .warpNavigationBarButton() // applies default style
    .warpNavigationBarButton(style: .primary) // applies primary style with prominent liquid glass style
```

### SwiftUI

Set up style appearance for Button
```swift
Button(action: {
     dismissView()
}) {
     Image(systemName: "xmark") // You should have use Warp icon instead
}
    .warpNavigationBarButton() // applies default style
    .warpNavigationBarButton(style: .primary) // applies primary style with prominent liquid glass style
```


AndreyMomot (commented):
Again, imho we shouldn't aim to merge it until we done with first step

Sources/Components/Native/NavigationBar/NavigationBar+Style.swift
```diff
@@ -0,0 +1,59 @@
+//
+// Created by Łukasz Śliwa on 26/02/2026.
+//
+
+import UIKit
+
+extension UINavigationBar {
```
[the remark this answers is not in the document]
luk-m-s: I'm not really convinced about that - Warp doesn't include backward compatibility + quirks when you transitioning between old (opaque) and new (transparent) navigation bars.

I would used this extensions as injections point for Warp, never views could use Warp modifiers directly when old ones could benefit from them via this extensions + hacks for seamless animation stuff 🤔 

This will be gradual change I think.

Sources/Components/Native/NavigationBar/UIBarButtonItem+Style.swift
```diff
@@ -0,0 +1,58 @@
...
+            tintColor = tint
+            setTitleTextAttributes([.foregroundColor: Warp.UIColor.token.text], for: .normal)
+            setTitleTextAttributes([.font: Warp.Typography.body.uiFont], for: .normal)
+            setTitleTextAttributes([.font: Warp.Typography.body.uiFont], for: .highlighted)
+        case .primary:
+            self.style = .prominent
+            tintColor = primaryTint
+            setTitleTextAttributes([.foregroundColor: Warp.UIColor.token.text], for: .normal)
+            setTitleTextAttributes([.font: Warp.Typography.body.uiFont], for: .normal)
+            setTitleTextAttributes([.font: Warp.Typography.body.uiFont], for: .highlighted)
```
AndreyMomot: `setTitleTextAttributes(_:for:)` replaces the entire attribute dictionary for a control state — it does not merge. The three sequential calls overwrite each other, so `.normal` ends up with only `.font` and loses `.foregroundColor` (the title falls back to the system tint).

Same bug in both `.default` and `.primary` branches.

Suggested fix — combine into one call per state:

```swift
setTitleTextAttributes(
    [.foregroundColor: Warp.UIColor.token.text,
     .font: Warp.Typography.body.uiFont],
    for: .normal
)
setTitleTextAttributes(
    [.foregroundColor: Warp.UIColor.token.text,
     .font: Warp.Typography.body.uiFont],
    for: .highlighted
)
```

A small XCTest asserting `titleTextAttributes(for: .normal)` contains both keys after `.warpNavigationBarButton(...)` would prevent regressions.

Sources/Components/Native/NavigationBar/NavigationBar+Style.swift
```diff
@@ -0,0 +1,127 @@
...
+            .font: Warp.Typography.title6.uiFont
+        ]
+
+        let backImage = Warp.Icon.chevronLeft.uiImage
+        appearance.setBackIndicatorImage(backImage, transitionMaskImage: backImage)
+
+        let plainButtonAppearance = UIBarButtonItemAppearance(style: .plain)
+        plainButtonAppearance.normal.titleTextAttributes = [
+            .foregroundColor: Warp.UIColor.token.text,
+            .font: Warp.Typography.title4.uiFont
```
AndreyMomot: Bar button typography mismatch between the two APIs in this PR:

- Static `warpLiquidGlassStyle()` (this file): `plainButtonAppearance.normal.titleTextAttributes` → `title4.uiFont`
- Per-button `UIBarButtonItem.warpNavigationBarButton()`: `body.uiFont`

Both style the same UI element. `body` matches the per-button modifier and the conventional sizing for bar button titles — `title4` looks too large here.

Sources/Components/Native/NavigationBar/ToolbarItem+Style.swift
```diff
@@ -0,0 +1,146 @@
...
+            case .default:
+                SwiftUI.Button(action: action) {
+                    buttonContent
+                }
+                .tint(colorProvider.token.icon)
+            case .primary:
+                SwiftUI.Button(action: action) {
+                    buttonContent
+                }
+                .buttonStyle(.borderedProminent)
```
AndreyMomot: `Warp.ToolbarItem(style: .primary)` uses `.borderedProminent`, while the sibling `Button.warpNavigationBarButton(style: .primary)` uses `.glassProminent`. On iOS 26 these render differently: `.borderedProminent` is a solid filled pill, `.glassProminent` is the Liquid Glass treatment. For consistency with the Button modifier and to actually deliver Liquid Glass on iOS 26+, this should be `.glassProminent` (gated to iOS 26 if you want a pre-26 fallback).

AndreyMomot (approved):
Took a closer look at the PR alongside what we have in our app (Drive + Extensions modules). Overall direction looks right - same Warp tokens, same layered API (global + per-instance), same .default/.primary shape we'd land on too. A few things to flag before this is something we can adopt cleanly:

  1. iOS 18 support. Every API is gated iOS 26 only and no-ops below. We're on iOS 18+ and our equivalents apply Warp styling on both - without that, iOS 18 users get system styling. Worth a fallback path.
  2. SwiftUI ↔ UIKit are fully parallel (three separate style enums, no shared underlying mechanism). We bridge them so a mixed screen gets one configuration path - without a bridge, the two surfaces will drift.
  3. Production hardening absent - we hit (and have fixes for) these on Liquid Glass bars: hosting controller .systemBackground flashing through transparent bars during push, fatal safe-area feedback loop at large Dynamic Type when extendedLayoutIncludesOpaqueBars is set mid-transition, container VC hierarchies where navigationItem lives on the parent, and VoiceOver label composition for icon buttons. Happy to share the relevant code if useful.

  Suggested path: either Warp absorbs those edge cases upstream, or we keep thin Drive wrappers on top of Warp's primitives so we don't lose what we learned. Either works - just shouldn't be a naive drop-in.

Sources/Components/Native/NavigationBar/NavigationBar+Style.swift
luk-m-s: Yup, omission on my side, fixed now 👍

Sources/Components/Native/NavigationBar/ToolbarItem+Style.swift
luk-m-s: Fixed, and gated to iOS 26

Sources/Components/Native/NavigationBar/UIBarButtonItem+Style.swift
luk-m-s: Done, + added some unit tests for that

luk-m-s:
I think the real problem is misunderstanding. I don't treat Drive + Extensions in adversary manner. To me they should be complementary.

1. In Warp we add Liquid Glass support with forward-only thinking. AFAIK we don't have unified designs for navigation/tabbar < iOS 26. So if we want to backport Drive+Extensions styling for them into Warp, sure - but for what value? Doing that we only add tech debt into design system which will be dropped when iOS 18 lost support (probably next year).
2. I unify enums, yep, maybe from a type system point of view it is better to keep them separate per component but that's a cosmetic change. Not sure what you mean by bridging SwiftUI/UIKit though - only official bridge is via UIHostingController/UIViewRepresentable and because currently SwiftUI is backed by UIKit backend some other APIs could affect both of them (like UIAppearanceProxy). But this is not guaranteed by Apple in any way 🤷 
3. Same as in point 1. I don't think Warp design system is a place for such code (with exception of VoiceOver label composition - can I take a look 👀 ?) especially when Warp isn't part of monorepo - animation glitches, combining liquid glass screens with non-liquid glass screens in seamless way, that kind of stuff. This code is important from POV of NMP app but is it really important from POV of design system? I think that approach simply slows down fixing speed :|

Summary: I think the real problem here is that you suppose that this change in Warp want to replace Drive/UIKit+Extensions fully. I don't think this is the case (at least not for near future :P )- we still have to support iOS 18, we still have a ton of screens that don't fully migrated to liquid glass. Liquid Glass still have some bugs inherent to itself. With this change I only want to introduce couple of shared methods which should handle Liquid Glass styling in unified way for  >= iOS 26 systems. I want to integrate them within Drive+Extensions for now. In future we can revisit if we still need Drive wrappers but then it should be a simple refactor :)

Sources/Components/Native/NavigationBar/ToolbarItem+Style.swift
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @ViewBuilder
+    private func buttonContent(iconColor: Color? = nil) -> some View {
+        HStack(spacing: Warp.Spacing.spacing100) {
+            if let icon {
+                Warp.IconView(icon, size: .default, color: iconColor)
+            }
+            if let title {
+                Warp.Text(title, style: .body)
```
AndreyMomot: `Warp.Text` falls back to `theme.token.text` when no color is passed (`Text.swift:72`), so primary buttons render the title in default text color on top of `buttonPrimaryBackground`. Existing Warp primary buttons use inverted text for contrast. Did you verify the title is readable in both the `.glassProminent` and `.borderedProminent` paths?

Sources/Components/Native/NavigationBar/UIBarButtonItem+Style.swift
```diff
@@ -0,0 +1,56 @@
...
+            setTitleTextAttributes([
+                .foregroundColor: Warp.UIColor.token.text,
+                .font: Warp.Typography.body.uiFont
+            ], for: .normal)
+            setTitleTextAttributes([
+                .foregroundColor: Warp.UIColor.token.text,
+                .font: Warp.Typography.body.uiFont
+            ], for: .highlighted)
+        case .primary:
+            self.style = .prominent
```
AndreyMomot, line 43: `.default` doesn't reset `self.style`, so a button that was first set to `.primary` (-> `.prominent`) and then re-applied with `.default` keeps `.prominent`. Either set `self.style = .plain` in the `.default` branch, or document this as one-shot only.

Sources/Components/Native/NavigationBar/Button+Style.swift
```diff
@@ -0,0 +1,42 @@
...
+                      .tint(Warp.Color.token.icon)
+                      .font(Warp.Typography.body.font)
+                case .primary:
+                    self
+                      .buttonStyle(.glassProminent)
+                      .tint(Warp.Color.buttonPrimaryBackground)
+                      .font(Warp.Typography.body.font)
+                }
+            } else {
+                self
```
AndreyMomot, line 28: The package targets iOS 17+, but this returns `self` on iOS < 26 with no font or tint - apps on iOS 17-25 get a bare system button. `WarpToolbarButton` does provide a `.borderedProminent` fallback, so it's inconsistent across the two SwiftUI entry points. Either give pre-26 a tint+font baseline here too, or mark the method `@available(iOS 26.0, *)` so callers see the constraint at compile time instead of a silent no-op.

AndreyMomot:
👍 thanks for explanation, not overall idea is much clearer to me 

Regarding Warp + ios26 understandable, we can use fallback to ios26< in monorepo but for this we should mark all Warp's ios26-only-related features as ios26 only, Example [here](https://github.com/warp-ds/warp-ios/pull/199#discussion_r3497820468)

AndreyMomot (approved):

AndreyMomot:
Also to avoid people use non-Warp existed solutions after we merge this I'd recommend to update docs + skills for ios26 and mention these changes there

Sources/Components/Native/NavigationBar/ToolbarItem+Style.swift
luk-m-s: 🤔 Good catch

luk-m-s: Good catch 👀

Sources/Components/Native/NavigationBar/UIBarButtonItem+Style.swift
luk-m-s: This is the place when SwiftUI (ToolbarItem) and UIKit starts to diverge :D On UIKit bordered and done style are deprecated but for SwiftUI there are still available 🤔 So yeah - for UIBarButtonItem I change it to `.plain` as other styles are deprecated

Not so sure about ToolbarItem counterpart though, in that place I don't force `buttonStyle` in default

Sources/Components/Native/NavigationBar/ToolbarItem+Style.swift
luk-m-s: Fixed

Sources/Components/Native/NavigationBar/Button+Style.swift
luk-m-s: I make all this method iOS 26 only, so no-one accidentally used it for older stuff :)

PR #202, opened 2026-04-21
Update components to SwiftUI `@Environment` theme injection

luk-m-s:
### Why
  This in future make easier to eliminate global mutable state (`Warp.Color`, `Warp.Token`) for strict concurrency compliance

  ### What
  Refactored 30+ components to use `@Environment(\.warpTheme)` instead of global singleton access:

  **Example**
  ```swift
  // Before
 Warp.Theme = .finn
SomeComponentView()

  // After
SomeComponentView()
    .warpTheme(.finn) // No global var here, propagate automatically down the view hierarchy
```

### DIFF
--- a/SampleApp/Blocket/Warp_iosBlocket.swift
+++ b/SampleApp/Blocket/Warp_iosBlocket.swift
@@ -6,13 +6,15 @@ import FirebaseCore
 struct Wrap_iosBlocket: App {
 
     init() {
-        Warp.Theme = .blocket
+        // Register fonts for the Blocket theme
+        try? Warp.Typography.registerFonts(for: .blocket)
         FirebaseApp.configure()
     }
-    
+
     var body: some Scene {
         WindowGroup {
             ContentView()
+                .warpTheme(.blocket)
         }
     }
 }
--- a/SampleApp/Blocket/Warp_iosBlocket.swift
+++ b/SampleApp/Blocket/Warp_iosBlocket.swift
@@ -7,7 +7,7 @@ struct Wrap_iosBlocket: App {
 
     init() {
         // Register fonts for the Blocket theme
-        try? Warp.Typography.registerFonts(for: .blocket)
+        try! Warp.Typography.registerFonts(for: .blocket)
         FirebaseApp.configure()
     }
 
--- a/Sources/Border/Border.swift
+++ b/Sources/Border/Border.swift
@@ -8,7 +8,8 @@ extension Warp {
         /// Border radius of 2 points. Typically used for very small rounded corners.
         public static let borderRadius25: CGFloat = 2
         /// Border radius of 4 points. A slightly larger rounded corner for small elements.
-        public static let borderRadius50: CGFloat = 4
+        public static let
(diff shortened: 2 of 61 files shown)

### CODE REVIEW

File: SampleApp/Blocket/Warp_iosBlocket.swift [hunk 1]
```diff
@@ -6,13 +6,15 @@ import FirebaseCore
 struct Wrap_iosBlocket: App {
 
     init() {
-        Warp.Theme = .blocket
+        // Register fonts for the Blocket theme
+        try? Warp.Typography.registerFonts(for: .blocket)
```
MarcinGie: I'm new here so it's just wondering - is a silent failure fine here? or would it be beneficial to have some error handing in place?

Sources/Border/Border.swift
```diff
@@ -8,7 +8,8 @@ extension Warp {
         /// Border radius of 2 points. Typically used for very small rounded corners.
         public static let borderRadius25: CGFloat = 2
         /// Border radius of 4 points. A slightly larger rounded corner for small elements.
-        public static let borderRadius50: CGFloat = 4
+        public static let
+            borderRadius50: CGFloat = 4
```
MarcinGie: Is the new line intentional here? ;)

Sources/Components/Modal/ModalViewModifier.swift
```diff
@@ -115,16 +115,29 @@ extension Warp {
     }
 }
 
+/// A workaround to make fullScreenCover backgrounds transparent on iOS 15.
```
MarcinGie: Do we still need to support iOS 15?

SampleApp/Blocket/Warp_iosBlocket.swift
luk-m-s: Hmm, after all this is demo app used mostly by designers - it doesn't mean to be used in any prod env 🤔 Probably force unwrap will be better here - instant feedback :P

Sources/Border/Border.swift
luk-m-s: 🤷

MarcinGie (approved):
Left some small questions, but LGTM 👍

SampleApp/Blocket/Warp_iosBlocket.swift
luk-m-s: Done.

Sources/Border/Border.swift
luk-m-s: Fixed.

SampleApp/Blocket/Warp_iosBlocket.swift
luk-m-s: I've added additional test case for checking Warp fonts registering, just for case... ;)

Sources/Components/Modal/ModalViewModifier.swift
luk-m-s: I bump version to iOS 17 and remove this legacy code

osanoj (approved):

PR #205, opened 2026-04-30
Add semantic spacing values

Breiby:
# Why?

Screen level spacing isn't something we should to think about when receiving a design, it should generally stay consistent on a screen-to-screen basis.

# What?

Add semantic spacing values for a few common types of spacing. There are more, but starting small. Think of the PR as a suggestion/request.


robinsalehjan (approved):

luk-m-s (approved):
👍

PR #206, opened 2026-05-07
WARP-1341 Center content of empty state views

luk-m-s:
## Why

Without centering the empty state views behaving erratically in Aurora app

## What

Use `ViewThatFits` component to dynamically switch between scroll view wrapped and raw state view content (so simultaneously we get centered if fit and scrollable component otherwise)


charlieeddus (approved):
I like it!

PR #208, opened 2026-05-19
WARP-1343 Create snackbar component

luk-m-s:
## Why

We need toast like component which supports buttons and looks more `glassy`

## What

Create `SnackbarView` component with API similar to toast but with modern look.

### Basic Snackbar (no action)

```swift
  @State private var showSnackbar = false
  
  var body: some View {
      VStack {
          Button("Save") { showSnackbar = true }
      }
      .warpSnackbar(
          type: .positive,
          title: "Changes saved successfully",
          isPresented: $showSnackbar
      )
  }
```

### Snackbar with Inline Action Button

```swift
  @State private var showSnackbar = false
  
  var body: some View {
      VStack {
          Button("Delete") { showSnackbar = true }
      }
      .warpSnackbar(
          type: .negative,
          title: "Item deleted",
          action: Warp.Snackbar.Action(title: "Undo") {
              restoreItem()
          },
          duration: .long,
          isPresented: $showSnackbar
      )
  }
```

### Snackbar with Long Action Button

```swift 

  @State private var showSnackbar = false

  var body: some View {
      VStack {
          Button("Retry") { showSnackbar = true }
      }
      .warpSnackbar(
          type: .warning,
          title: "Connection lost",
          longAction: Warp.Snackbar.Action(title: "Retry Connection") {
              reconnect()
          },
          duration: .infinite,
          showCloseButton: true,
          isPresented: $showSnackbar
      )
  }
```

## Preview

See Snackbar snapshots


Sources/Components/Snackbar/Snackbar.swift
```diff
@@ -0,0 +1,298 @@
...
+            .cornerRadius(snackbarCornerRadius)
+            .overlay(
+                RoundedRectangle(cornerRadius: snackbarCornerRadius)
+                    .stroke(SwiftUI.Color.white.opacity(0.15), lineWidth: 1)
+            )
+            .transition(AnyTransition.move(edge: .bottom).combined(with: .opacity))
+            .task {
+                do {
+                    try await Task.sleep(nanoseconds: duration.timeInNanoseconds)
+                    isPresented.toggle()
```
AndreyMomot: Both `.task` and `.onTapGesture` use `isPresented.toggle()` to dismiss. If the binding changes externally (e.g. parent hides the snackbar before the timer fires), `toggle()` would flip it back to `true`. Using `isPresented = false` is safer - dismissal should be idempotent.

Sources/Components/Snackbar/SnackbarDuration.swift
```diff
@@ -0,0 +1,31 @@
...
+        case custom(interval: TimeInterval)
+
+        var timeInNanoseconds: UInt64 {
+            switch self {
+            case .short:
+                5_000_000_000 // 5 seconds
+            case .long:
+                20_000_000_000 // 20 seconds
+            case .infinite:
+                6_000_000_000_000
```
AndreyMomot: `.infinite` maps to `6_000_000_000_000` ns (~100 minutes), so it will still auto-dismiss eventually. The doc says "requires user interaction to dismiss" - the `.task` should skip `Task.sleep` entirely for `.infinite` instead of faking it with a large number.

Tests/WarpSnapshotTests/Components/SnackbarSnapshotTests.swift
```diff
@@ -0,0 +1,106 @@
...
+
+    enum ActionMode {
+        case none
+        case inline
+        case long
+    }
+
+    @Test(arguments: Self.allArgumentsCombined)
+    func snapshotAllSnackbars(brand: Warp.Brand, type: Warp.SnackbarType, showCloseButton: Bool, actionMode: ActionMode) {
+        let snapshotName = [
```
AndreyMomot, line 27: `showCloseButton` is in `allArgumentsCombined` but not in `snapshotName` - the `true`/`false` pairs will produce the same filename and overwrite each other 👀 Also, since the test uses `.infinite` duration for all cases, `Snackbar.swift:243` (`showCloseButton || duration == .infinite`) always forces the close button on, making `showCloseButton: false` untestable with this setup.

Sources/Components/IconView/PaletteIconView.swift
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+extension Warp.PaletteIconView {
+    /// Applies a custom foreground color to the `IconView`.
+    ///
+    /// - Parameter color: The custom color to apply to the icon.
+    /// - Returns: A new `IconView` with the updated color.
+    public func foregroundColor(_ color: Color) -> Self {
+        Warp.PaletteIconView(icon, size: size, color: color)
```
AndreyMomot: `foregroundColor(_:)` creates `PaletteIconView(icon, size: size, color: color)` without passing `paletteColor` through - chaining `.foregroundColor(...)` after init with a custom `paletteColor` silently reverts the target color back to `.black`.

Sources/Components/IconView/WarpPaletteIconManager.swift
```diff
@@ -0,0 +1,58 @@
...
+
+    func recolorRasterizedImage(
+        icon: Warp.Icon,
+        targetColor: UIColor = .black,
+        newColor: UIColor,
+        size: Warp.IconSize = .default,
+        tolerance: Float = 0.15
+    ) -> UIImage? {
+
+        let cacheKey = "\(icon.assetName)_\(size.value)x\(size.value)_\(newColor.hashValue)" as NSString
```
AndreyMomot: The cache key uses `newColor.hashValue` which isn't stable and can collide for different colors. It also doesn't include `targetColor` or `tolerance` - two calls with different `targetColor` but matching `icon`/`size`/`newColor` would return the wrong cached image. Consider building the key from resolved RGBA components of both colors.

Sources/Components/Snackbar/Snackbar.swift
```diff
@@ -0,0 +1,298 @@
...
+            SwiftUI.Button(action: action.handler) {
+                Text(action.title, style: .bodyStrong)
+                    .foregroundColor(type.textColor(from: colorProvider))
+            }
+        }
+
+        private var closeButton: some View {
+            Group {
+                if showCloseButton || duration == .infinite {
+                    Warp.IconView(.close, size: .small, color: colorProvider.token.iconInvertedStatic)
```
AndreyMomot: The close icon is `Warp.IconView(.close, ...)` but not wrapped in a `Button` - no distinct tap target, only the whole-snackbar `onTapGesture`. If there's an action button present, tapping the close icon triggers the parent gesture instead of a specific dismiss. Wrapping it in a `Button` would also give VoiceOver users a separate close affordance.

Sources/Components/Snackbar/Snackbar.swift
```diff
@@ -0,0 +1,298 @@
...
+                RoundedRectangle(cornerRadius: snackbarCornerRadius)
+                    .fill(.ultraThinMaterial)
+            )
+            .cornerRadius(snackbarCornerRadius)
+            .overlay(
+                RoundedRectangle(cornerRadius: snackbarCornerRadius)
+                    .stroke(SwiftUI.Color.white.opacity(0.15), lineWidth: 1)
+            )
+            .transition(AnyTransition.move(edge: .bottom).combined(with: .opacity))
+            .task {
```
AndreyMomot, line 178: The auto-dismiss in `.task` is not wrapped in `withAnimation`, but the manual tap dismiss on line 182 is. Timer-based dismiss will snap away without the slide-out transition.

Sources/Components/Snackbar/Snackbar.swift
```diff
@@ -0,0 +1,298 @@
...
+                do {
+                    try await Task.sleep(nanoseconds: duration.timeInNanoseconds)
+                    isPresented.toggle()
+                } catch {} // Handle cancellation if needed
+            }
+            .onTapGesture {
```
AndreyMomot, line 200: `.onTapGesture` on the outer `Group` will compete with the action `Button` - tapping the action button could also trigger the dismiss gesture. Did you verify the action handler fires correctly when an action button is present?

Sources/Components/Snackbar/Snackbar.swift
```diff
@@ -0,0 +1,298 @@
...
+        }
+    }
+}
+
+extension Warp.SnackbarType {
+    fileprivate func backgroundColor(from colorProvider: ColorProvider) -> Color {
+        colorProvider.tooltipBackgroundStatic
+    }
+
+    fileprivate func subtleBorderColor(from colorProvider: ColorProvider) -> Color {
```
AndreyMomot: `subtleBorderColor(from:)` is defined here but never called anywhere in this PR

Sources/Components/IconView/PaletteIconView.swift
```diff
@@ -0,0 +1,96 @@
...
+        private let size: Warp.IconSize
+        /// The color of the icon from the palette which will be tinted. If nil, defaults to black.
+        private let paletteColor: Color?
+        /// The color of the icon. If nil, defaults to theme's icon color.
+        private let color: Color?
+
+        /// The current theme from the environment.
+        @Environment(\.warpTheme) private var theme
+
+        var paletteIconManager: WarpPaletteIconManager = .shared
```
AndreyMomot, line 20: `paletteIconManager` is `internal var` but looks like an implementation detail - should this be `private`?

Sources/Components/IconView/PalletteIconView.swift
```diff
@@ -0,0 +1,96 @@
+import SwiftUI
+
+extension Warp {
+    /// A view that renders an palette icon with the specified size and color.
```
AndreyMomot: Doc comment says "an palette icon" and references `IconView` in three places - should be "a palette icon" and `PaletteIconView`. Also the filename itself has an extra 'l' (`Pallette` vs `Palette`).

Sources/Components/Snackbar/Snackbar.swift
luk-m-s: Done.

Sources/Components/Snackbar/SnackbarDuration.swift
luk-m-s: Fixed in task logic + make `.infinite` equal to UInt64.max

Sources/Components/IconView/WarpPaletteIconManager.swift
luk-m-s: I replaced this with already existing method `hexString`. It covers only `RGB` without alpha channel but because this is internal helper object I think it's acceptable - for palette tinting you probably don't want to mess with alpha channel anyway.

Sources/Components/Snackbar/Snackbar.swift
luk-m-s: 👍

Sources/Components/Snackbar/Snackbar.swift
luk-m-s: Strange - it works in demo app with transition 🤷‍♂️

Sources/Components/Snackbar/Snackbar.swift
luk-m-s: Button fires correctly - unless Apple completely abandon UIKit as a part of under the hood imp. detail it should behave similarly to UIKit (so responder chain always delegates hit touch to its subviews).

Sources/Components/Snackbar/Snackbar.swift
luk-m-s: Done.

Sources/Components/IconView/PalletteIconView.swift
luk-m-s: Typo.

Sources/Components/IconView/PaletteIconView.swift
luk-m-s: This is intentional, this is dependency, just in case left myself with the possibility to replace it with some non-shared instance (testing? dunno)

Sources/Components/IconView/PaletteIconView.swift
luk-m-s: Fixed.

Tests/WarpSnapshotTests/Components/SnackbarSnapshotTests.swift
luk-m-s: Good catch 👍 I've added `showCloseButton` to the snapshot description and switch `duration` from `infinite` to `long`. The auto-dismissal logic shouldn't affect snapshot test but now you can have properly tested show/hide close button layout.

Sources/Components/Snackbar/SnackbarDuration.swift
```diff
@@ -0,0 +1,31 @@
...
+        var timeInNanoseconds: UInt64 {
+            switch self {
+            case .short:
+                5_000_000_000 // 5 seconds
+            case .long:
+                10_000_000_000 // 10 seconds
+            case .infinite:
+                UInt64.max
+            case .custom(let interval):
+                UInt64(interval * 1_000_000_000) // Convert time interval to nanoseconds
```
AndreyMomot: `UInt64(interval * 1_000_000_000)` will trap at runtime for negative, NaN, or very large `TimeInterval` values. Since `custom(interval:)` is a public API accepting arbitrary `TimeInterval`, a consumer passing a negative value gets a crash. Consider clamping: `UInt64(max(0, interval) * 1_000_000_000)` or validating in the init.

Sources/Icon/Icon.swift
```diff
@@ -308,6 +312,13 @@ extension Warp {
                 .resizable()
         }
 
+        /// Returns a `UIImage` for the corresponding icon, loaded from the asset catalog, without applying template rendering mode.
+        var rawUIImage: UIImage? {
+            let image = UIImage(named: assetName, in: .module, compatibleWith: nil)
+            image?.accessibilityLabel = localization
```
AndreyMomot, line 319: `image?.accessibilityLabel = localization` mutates a `UIImage` returned by `UIImage(named:in:compatibleWith:)`. UIKit caches these image objects, so this mutation leaks the accessibility label to every other consumer of the same asset. Set the label on the SwiftUI `Image` view layer instead, or create a copy of the UIImage before mutating.

Sources/Components/IconView/PaletteIconView.swift
```diff
@@ -0,0 +1,97 @@
...
+                    size: Warp.IconSize = .default,
+                    paletteColor: Color? = nil,
+                    color: Color? = nil) {
+            self.icon = icon
+            self.size = size
+            self.paletteColor = paletteColor
+            self.color = color
+        }
+
+        // Tint image and cache it
```
AndreyMomot, line 39: `tintedImage` is a computed property evaluated every time `body` renders, and on a cache miss it does a synchronous CoreImage round-trip on the main thread (the comment even says "halts the main thread"). For a snackbar animating in, this is exactly the wrong time to hitch. Consider moving the recoloring into a `@State` populated from `.task`, or pre-rendering the four snackbar icons at init time.

Sources/Components/Snackbar/Snackbar.swift
```diff
@@ -0,0 +1,325 @@
...
+            self._isPresented = isPresented
+        }
+
+        public var body: some View {
+            Group {
+                if longAction != nil {
+                    longActionContentView
+                } else {
+                    actionContentView
+                }
```
AndreyMomot, line 162: `maxWidth: horizontalSizeClass == .regular ? .infinity : 420` - this gives iPad full width while capping iPhone at 420pt. Typically it's the other way around: constrain the snackbar on iPad so it doesn't stretch across the entire screen. Did you verify this matches the intended design spec?

Sources/Components/Snackbar/Snackbar.swift
```diff
@@ -0,0 +1,325 @@
+import Foundation
+import SwiftUI
+
+extension Warp {
+
+    private static let snackbarCornerRadius: Double = 24
```
AndreyMomot, line 7: Several hardcoded visual constants in a public Warp component: `snackbarCornerRadius = 24`, `.opacity(0.9)`, `SwiftUI.Color.white.opacity(0.15)`, `.padding(.leading, 8)`. Since this is a design system component, these should ideally reference Warp tokens (border radius, spacing, color tokens) so they stay in sync when design tokens are updated.

AndreyMomot (approved):
LGTM with some minor comments below

Sources/Components/IconView/PaletteIconView.swift
luk-m-s: This is on purpose.  Moving it to async main thread breaks runloop continuity. For the reasons unknown this removes animation layer in SwiftUI component. If you animate Snackbar component and it spawns image tinting then it looks like icon is already **inplace**. I don't find any solution to this quirk other than make it synchronous.

This component is not for public consumption anyway so it should be used with care. Icons are 24x24px, rendering them on main thread should not create any stutter on modern devices.

Sources/Components/Snackbar/Snackbar.swift
luk-m-s: Unfortunately for now we don't have corner radius token matching Liquid glass :/ This will change in the future -  I've added `todo` comment. 

Opacity is related to glass material (opaque colors don't work), same for outlier. 
Padding should be fixed now.

Sources/Icon/Icon.swift
luk-m-s: Huh? Looks like another Apple API issue (if it's true that UIImage instances share it's internal state) :D 

I've added this method only to get icon image without template rendering for palette tinting. It doesn't suppose to be public - I say this could be safely ignored ;)

AndreyMomot (approved):

Sources/Components/Snackbar/SnackbarDuration.swift
luk-m-s: Added

Sources/Components/Snackbar/Snackbar.swift
luk-m-s: Yeah, I leave with `maxWidth = 500` This prevents full width on larger iPads but allow it on iPhones

PR #216, opened 2026-06-23
WARP-1363 Create  Warp ConfirmationDialog component

luk-m-s:
## What 

Adds a new ConfirmationDialog component that wraps SwiftUI's native `.confirmationDialog()` with Warp styling.

  ## Usage
  
  ```swift
  .warpConfirmationDialog(
      title: "Choose an action",
      message: "Select what you'd like to do.",
      actions: [
          Action(title: "Share") { /* handle share */ },
          Action(title: "Delete", style: .destructive) { /* handle delete */ }
      ],
      isPresented: $showDialog
  )
```

  The confirmation dialog appears as a native iOS sheet from the bottom with Warp typography and color tokens. Cancel button is provided automatically.

### CODE REVIEW

File: SampleApp/ContentView.swift [hunk 1]
```diff
@@ -41,6 +41,7 @@ struct ContentView: View {
                     Warp.Text("Warp Components", style: .title3)
                     Divider()
 
+                    row("Confirmation Dialog", destination: ConfirmationDialogView())
```
MarcinGie: Can this be moved to preserve the alphabetical order?

SampleApp/ContentView.swift
```diff
@@ -41,6 +41,7 @@ struct ContentView: View {
                     Warp.Text("Warp Components", style: .title3)
                     Divider()
```
MarcinGie: Food for thought: we have quite a list of components here, perhaps, **in the future** it would be useful to add a search bar that filters rows of this VStack to make selection faster?

MarcinGie (approved):
LGTM 👍

SampleApp/ContentView.swift
luk-m-s: Done

SampleApp/ContentView.swift
luk-m-s: Added 😁

PR #217, opened 2026-06-25
Add Claude skill for importing Warp icons from web/elements repository

luk-m-s:
Adds a Claude Code skill that documents the process for importing regular icons from the warp-ds/elements web repository into the Warp iOS library.

 ## What

  - Step-by-step import process from web repository to iOS
  - Asset catalog structure and configuration
  - Swift enum integration pattern
  - Translation format for all 5 supported languages (DA, EN, FI, NB-NO, SV)
  - Unit test validation workflow
  - Troubleshooting guide for common issues

 ## Usage
  
  ```bash
  /skill warp-icon-import
```

  The skill guides developers through the complete workflow including creating the asset catalog imageset, adding Swift enum cases, configuring translations, and running unit tests to verify the import.

### CODE REVIEW

File: .claude/skills/warp-icon-import.md [hunk 1]
```diff
@@ -0,0 +1,264 @@
...
+
+Add case to the `Icon` enum:
+```swift
+public enum Icon: String, CaseIterable {
+    // ... existing cases
+    case {camelCaseName}  // e.g., case activeAds
+}
+```
+
+### 4. Add Accessibility Title Mapping
```
MarcinGie: Nice work on the overall structure here! One thing to flag — the actual property in `Icon.swift` is `var localization: String`, not `var title: String`. The heading says `"Add Accessibility Title Mapping"` and the code block shows public `var title: String`, but anyone following this will be looking for a property that doesn't exist.

_Suggestion_: rename the heading to `"Add Accessibility Localization Mapping"` and update `line 78` to `public var localization: String`.

.claude/skills/warp-icon-import.md
```diff
@@ -0,0 +1,264 @@
...
+
+In the same file, add the localized title mapping:
+```swift
+public var title: String {
+    switch self {
+    // ... existing mappings
+    case .{camelCaseName}: return Warp.Strings.icon{PascalCaseName}.localized
```
MarcinGie, line 85: This line references `Warp.Strings.icon{PascalCaseName}.localized`, but the skill doesn't explain where `Warp.Strings` comes from or how it gets generated. If it's auto-generated from the xcstrings file (e.g. via a build step), it'd be really helpful to mention that — otherwise someone following the guide might wonder why `Warp.Strings.iconNewIcon` doesn't exist yet and think something went wrong. Even a one-liner like `"This property is auto-generated from the translation keys added in step 5"` would go a long way.

.claude/skills/warp-icon-import.md
```diff
@@ -0,0 +1,264 @@
...
+| camelCase | activeAds | Swift enum case, translation key suffix |
+
+## Checklist
+
+- [ ] Download SVG from warp-ds/elements repository
+- [ ] Rename to PascalCase (e.g., active-ads.svg → ActiveAds.svg)
+- [ ] Create .imageset folder in RegularIcons/
+- [ ] Add SVG and Contents.json to .imageset
+- [ ] Add enum case to Icon.swift (camelCase)
+- [ ] Add title mapping in Icon.swift
```
MarcinGie: Same naming mismatch as step 4 — this says "Add title mapping" but the property is actually called `localization`. Small thing, but keeping the terminology consistent with the codebase will make the checklist easier to follow.

.claude/skills/warp-icon-import.md
```diff
@@ -0,0 +1,264 @@
...
+
+# Warp Icon Import
+
+Process for importing regular icons from the warp-ds/elements web repository into the Warp iOS library with corresponding accessibility translations.
+
+## Source Repository
+
+Icons are sourced from: https://github.com/warp-ds/elements
+
+Web icon location: `warp-ds/elements/components/icons/src/`
```
MarcinGie, line 14: This is a handy reference, but if the elements repo ever restructures, this path could silently become wrong and lead Claude (or a dev) down a dead end. Might be worth adding a note like "verify this path against the current repo structure" or just linking to the repo root and describing how to navigate to the icons folder. **Totally optional** — just thinking about long-term maintenance of the skill.

.claude/skills/warp-icon-import.md
```diff
@@ -0,0 +1,264 @@
...
+- [ ] Create .imageset folder in RegularIcons/
+- [ ] Add SVG and Contents.json to .imageset
+- [ ] Add enum case to Icon.swift (camelCase)
+- [ ] Add title mapping in Icon.swift
+- [ ] Add translations to Localizable.xcstrings (icon.title.{camelCase})
+- [ ] Verify all 5 languages have translations
+- [ ] **Run icon unit tests** (WarpTests/IconTests)
+- [ ] Verify all tests pass
+- [ ] Build project to verify no errors
+- [ ] Test icon displays correctly in SampleApp
```
MarcinGie, line 218: The checklist says "Test icon displays correctly in SampleApp" and the related files section references `IconsView.swift`, but it's not mentioned that newly added icons appear automatically because the Icon enum conforms to `CaseIterable`. A quick note like `"New icons appear automatically in the demo app since Icon is CaseIterable"` would save someone from looking for where to manually wire things up.

MarcinGie (approved):
Well-structured skill. Clear step-by-step process, good naming convention table, useful troubleshooting section, and a practical checklist. The PR description matches what's delivered. A few inaccuracies to fix mentioned in comments 👍

.claude/skills/warp-icon-import.md
luk-m-s: Done

.claude/skills/warp-icon-import.md
luk-m-s: 👍

.claude/skills/warp-icon-import.md
luk-m-s: 👍

PR #218, opened 2026-06-26
WARP-1358 Add UITabBar styling helpers (warpStyle, icon tinting, native badge)

osanoj:
## Summary

Adds Warp helpers for styling a UIKit `UITabBar` and its items with Warp tokens, so apps get consistent, brand-aware tab bars without hand-rolling appearance/badge code. Uses Apple's **native** `UITabBarItem` badge (styled via the appearance proxy) instead of a custom view, so it stays robust across iOS versions and renders correctly in the iOS 26 Liquid Glass tab bar.

## What's added

**Icon tinting** — `Sources/Components/Native/TabBar/UIImage+TabBarItem.swift`
- `Warp.TabBarItemState` (`.normal` / `.selected`)
- `UIImage.warpTabBarItemTinted(state:)` — tints a template tab icon with `Warp.UIToken.icon` / `Warp.UIColor.navbarIconSelected`.

**Bar appearance** — `UITabBar+Style.swift`
- `UITabBar.warpStyle()` (per-instance) + `UITabBar.warpConfigureAppearanceProxy()` (global): pre-iOS-26 opaque styling and iOS 26 Liquid Glass selected colours, the compact-size-class unselected tint with trait-change re-application, and native badge styling via the appearance proxy. Plus a `UITabBarController.warpStyle()` convenience.

**Native badge** — `UITabBarItem+WarpBadge.swift`
- `warpSetBadge(count:max:)` (caps at `"99+"`, clears on ≤ 0), `warpSetBadgeDot()`, `warpClearBadge()`, `warpConfigureImages(_:)`.
- Brand-red `backgroundNotification`, `textInvertedStatic` text, `detailStrong` font — applied through the appearance proxy. Localized VoiceOver labels (`Localizable.xcstrings`, Crowdin-synced).

**Tests & demo**
- Unit tests for the tinting and badge helpers.
- Cross-brand snapshot of a window-hosted styled tab bar (count + dot badge).
- Interactive **Bottom nav / Tab Bar** demo in the SampleApp (tab picker, add count/dot, remove).

## Notes
- Warp tokens only — no raw `UIColor`.
- Snapshot tests aren't part of `run-tests.yml` (unit-only); baselines are visual references.

## How to test
- SampleApp → Native Components → **Bottom nav / Tab Bar**: switch brand, add/remove count and dot badges.
- `xcodebuild -scheme WarpTests -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone Air' test`


Sources/Components/Native/TabBar/UITabBarItem+WarpBadge.swift
```diff
@@ -0,0 +1,68 @@
+// Sources/Components/Native/TabBar/UITabBarItem+WarpBadge.swift
+import UIKit
+
+public extension UITabBarItem {
```
luk-m-s, line 4: Not related to this file but maybe worth adding some SwiftUI modifiers/extensions also (i.e. TabView + Tab (like maybe only accepts Warp color, dunno if applicable :) ). I do sth similar in navigation bar.

Tests/WarpTests/Components/Native/TabBar/UIImageTabBarItemTintedTests.swift
```diff
@@ -0,0 +1,67 @@
+// Tests/WarpTests/Components/Native/TabBar/UIImageTabBarItemTintedTests.swift
+import Testing
+import UIKit
+@testable import Warp
+
+@Suite @MainActor
+struct UIImageTabBarItemTintedTests {
```
luk-m-s, line 7: Smells like snapshot test in units but it's a good workaround ;)

luk-m-s (approved):
Ok for me, maybe worth adding SwiftUI extensions/wrappers. 👍 

Also, I'm not sure if snapshots are really useful here - internal rendering looks broken for Liquid Glass for now :/

Tests/WarpTests/Components/Native/TabBar/UIImageTabBarItemTintedTests.swift
osanoj: Yes, you are probably right that those snapshots are not very useful 😄

Sources/Components/Native/TabBar/UITabBarItem+WarpBadge.swift
osanoj: You mean to prepare for a future when the main app will not have UIKit in the bottom? 🤔

luk-m-s: Is the Warp design system only purpose is to be utilized by the main app? 😆

AndreyMomot (approved):
LGTM 👍 


Two follow-ups (non-blocking):

- `warpConfigureAppearanceProxy()` is incomplete on iOS 18: backgroundColor / tintColor aren't proxy-settable on UITabBar, so apps that only call the proxy at launch won't get brand styling on pre-26 OS. 

- WarpTabBarBadgeStylingSnapshotTests sets Warp.Theme, mutates UITabBar.appearance() / UITabBarItem.appearance() (global, applies to every tab bar in the process), and creates a key window without tearing any of it down. Any future snapshot test that uses a UITabBar will inherit whichever brand styling this test left behind, and the "winner" depends on test ordering. A defer to reset the theme + proxies + release the window would isolate it.

PR #219, opened 2026-07-03
WARP-1403 Create Warp native `Alert` component

luk-m-s:
## What

  New `Alert` component wrapping SwiftUI's native `.alert()` modifier with Warp-typed actions. 

  ## Usage

  ```swift
  Text("Content")
      .warpAlert(
          title: "Delete item?",
          message: "This action cannot be undone.",
          actions: [
              Warp.AlertDialog.Action(title: "Delete", style: .destructive) { handleDelete() },
              Warp.AlertDialog.Action(title: "Cancel", style: .cancel) { }
          ],
          isPresented: $showAlert
      )
```

  Notes
  
  - Cancel button is explicit via .cancel style (unlike ConfirmationDialog where system provides it automatically)
  - Native alert title/message styling is system-controlled; button label styling follows Warp design system

### CODE REVIEW

File: Sources/Components/AlertDialog/AlertDialogViewModifier.swift [hunk 1]
```diff
@@ -0,0 +1,81 @@
+import SwiftUI
+
+extension Warp {
+
+    /// A view modifier that presents a native alert dialog with Warp-typed actions.
+    struct AlertDialogViewModifier: ViewModifier {
```
MarcinGie: The `ConfirmationDialogViewModifier` reads the theme via `@Environment(\.warpTheme)` and resolves colors through the colorProvider (see [ConfirmationDialogViewModifier.swift lines 13–16](https://github.com/warp-ds/warp-ios/blob/main/Sources/Components/ConfirmationDialog/ConfirmationDialogViewModifier.swift#L14)). This modifier skips that and accesses `Warp.Token` / `Warp.Typography` directly instead.

This means if a parent view overrides the theme with `.warpTheme()`, the alert won't pick it up — it'll always use the global theme. Probably fine in most cases today, but it breaks the contract the other native components set up. Worth aligning for consistency so consumers don't get surprised later.

Sources/Components/AlertDialog/AlertDialogViewModifier.swift
```diff
@@ -0,0 +1,81 @@
...
+                .alert(
+                    SwiftUI.Text(title),
+                    isPresented: $isPresented
+                ) {
+                    ForEach(actions) { action in
+                        SwiftUI.Button(role: action.style.buttonRole) {
+                            action.handler()
+                        } label: {
+                            SwiftUI.Text(action.title)
+                                .font(Warp.Typography.title4.font)
```
MarcinGie, line 33: Small heads-up: iOS `.alert()` fully controls how action buttons are rendered — it ignores `.font()` and `.foregroundStyle()` modifiers applied inside the actions builder. So the styling here won't have any visible effect at runtime.

I know `ConfirmationDialogViewModifier` has the same pattern (and confirmation dialogs also ignore it), so this is inherited rather than new.
Just flagging it so we're aware - we could either remove the dead styling or leave it with a short comment explaining why it's there (e.g. for documentation intent / future iOS changes). Up to you!

Sources/Components/AlertDialog/AlertDialog.swift
```diff
@@ -0,0 +1,70 @@
+import SwiftUI
+
+extension Warp {
+
+    /// A Warp wrapper around the native iOS alert dialog.
+    ///
+    /// `AlertDialog` provides data models for configuring a native alert dialog
+    /// with Warp-typed actions. Use the `.warpAlertDialog()` view modifier to present it.
```
MarcinGie: Minor doc mismatch — the comment says "Use the `.warpAlertDialog()` view modifier" but the actual public API is `.warpAlert()` (defined in `AlertDialogViewModifier.swift`, line 71). Could confuse someone relying on the docs to find the right call.

SampleApp/ContentView.swift
```diff
@@ -104,6 +103,8 @@ struct ContentView: View {
                     Warp.Text("Native Components", style: .title3)
                     Divider()
 
+                    row("Alert (Native)", destination: AlertDialogView())
```
MarcinGie: The "Alert (Native)" and "Confirmation Dialog" rows are hardcoded outside the filtered item lists, so they don't respond to the search bar.
Not a blocker by any means — just a small UX gap in the demo app. The Confirmation Dialog row had the same situation before this PR since it was in the unfiltered `componentItems`, so this is more of a pre-existing thing. Could be addressed separately if we want all demo rows to be searchable.

Sources/Components/AlertDialog/AlertDialogViewModifier.swift
luk-m-s: Fixed now 👍

Sources/Components/AlertDialog/AlertDialogViewModifier.swift
luk-m-s: I leave this with code-comment.

For me this looks like future-proof API, especially when native component allows you composition with @ViewBuilder but then magically `strips` it's style modifiers. Probably Apple don't allow customization now but they've could enable it in the future without breaking API 🤷 

Strange decisions from Apple side anyway.

SampleApp/ContentView.swift
luk-m-s: Fixed.

Sources/Components/AlertDialog/AlertDialog.swift
luk-m-s: Fixed.

MarcinGie (approved):
👍

PR #220, opened 2026-07-15
WARP-1410 Add sheet style modifiers

MarcinGie:
## Why

The Sheet component was listed as "Not started" in the component tracking sheet. The Figma spec defines it as style modifiers on the native iOS `.sheet()` rather than a custom Warp component.

## Scope note

**This PR is sheet-only.** An earlier revision also added pre-iOS-26 navigation bar APIs (`warpSolidStyle()`, `warpStyle()`, `warpNavigationBarSolid()`). Those have been removed after discussion: they made the Glass vs Solid version-dispatch decision inside Warp and applied it behind Drive's back, which conflicts with the "Drive API powers everything" decision for the global NMP navbar. Navigation bar styling stays with the consumer, and the broader Warp/Drive ownership question is being handled separately.

## What

`.warpSheetStyle(_:dragIndicator:background:)`, applied to the root content inside a `.sheet { }` closure.

- **Detents:** from `Warp.SheetStyle` - `.medium` (medium + large) or `.fullScreen` (large only).
- **Drag indicator:** a parameter defaulting to `.visible`, rather than being hardcoded per style. The Figma spec draws the grabber on the full-screen variant too, and Drive's call sites each decide for themselves.
- **Background:** `Warp.SheetBackground` - `.automatic`, `.solid`, `.color(_:)`, or `.none`.

Also adds a `warpNativeGlassEffectsEnabled` environment key, so hosts that can't rasterize native Liquid Glass (offscreen snapshot hosts) can opt out.

## What Glass and Solid actually mean

An earlier revision of this PR assumed Glass and Solid mapped to OS availability, with Solid as the pre-26 fallback. That was wrong, and @Adrian Dick corrected it: **Solid is a caller decision**, taken when translucency would hurt legibility over a busy backdrop or when the sheet should read as a focused task rather than an extension of the view behind it.

Measured on simulators, with no `presentationBackground` applied at all:

| OS | Detent | System renders |
|---|---|---|
| iOS 26.5 | `.medium` | Liquid Glass, backdrop visible through the sheet |
| iOS 26.5 | `.large`, reached by dragging | Fully opaque |
| iOS 26.5 | `.large`, as the only detent | Fully opaque |
| iOS 18.5 | any | Fully opaque |

So **iOS already switches between glass and opaque per detent**, without being asked. `.automatic` therefore stays out of the way wherever a glass state is reachable, and only supplies a colour where the presentation is opaque for its whole life:

- `.medium` keeps deferring to the system, so glass is preserved.
- `.fullScreen` now gets `surfaceElevated100`. It only has the large detent, so it is never translucent, and deferring there only meant taking Apple's colour instead of the brand's. Invisible in light mode, where both are `#ffffff`. Visible in dark mode, where the system uses `#1c1c1e` and the token is `#26262b` across all six brands.

**One case is deliberately left alone:** dragging a `.medium` sheet up to full height on iOS 26 still lands on Apple's colour. Overriding that would mean giving up glass at the medium detent, which is the appearance the spec asks for. Documented on the modifier; `.solid` is the way out for sheets that must stay on-brand at every height.

`SheetBackground.resolvedColor(token:style:supportsLiquidGlass:nativeGlassEffectsEnabled:)` takes availability as an injected parameter rather than reading `#available` internally, so both branches stay reachable from tests regardless of which OS the suite runs on.

## Usage

```swift
.sheet(isPresented: $showSheet) {
    NavigationStack {
        MyContentView()
            .navigationTitle("Title")
    }
    .warpSheetStyle(.medium)
}
```

## Tests

`Tests/WarpTests/Native/Sheet/SheetStyleModifierTests.swift`.

**TEST SUCCEEDED** on both ends of the supported range, 68 tests in 9 suites, 0 failures: iOS 18.5 (the package floor) and iOS 26.5. The `Finn` SampleApp scheme builds on both.

Snapshot tests aren't run here: they aren't part of CI, and a presented sheet can't be captured by the snapshot host at all, so a reference would be blank and would pass silently. The behaviour above was verified by presenting real sheets in a simulator and sampling the rendered pixels instead.


luk-m-s (approved):
Looks good 👍

SampleApp/Components/Native/SheetDemoView.swift
```diff
@@ -0,0 +1,214 @@
...
+                    }
+                }
+                ToolbarItem(placement: .topBarTrailing) {
+                    Button {
+                        isPresented = false
+                    } label: {
+                        Warp.IconView(.close, size: .default)
+                    }
+                    .accessibilityLabel("Close")
+                }
```
luk-m-s, line 184: `Warp.IconView` automatically add localized accessibility labels for icons. Here "Close" is hardcoded. As `Warp.Button` probably doesn't fit there maybe we could simply borrow accessibility label from Icon? Or directly call it from localizable strings? 🤔

Sources/Components/Native/Sheet/SheetStyleModifier.swift
```diff
@@ -0,0 +1,207 @@
...
+                        token: token,
+                        style: style,
+                        supportsLiquidGlass: Self.supportsLiquidGlass,
+                        nativeGlassEffectsEnabled: nativeGlassEffectsEnabled
+                    )
+                )
+        }
+
+        static var supportsLiquidGlass: Bool {
+            if #available(iOS 26.0, *) {
```
luk-m-s, line 132: I think this method could have wider use than this component, can we extract it in some shared space (maybe under the Warp)?

Tests/WarpTests/Native/Sheet/SheetStyleModifierTests.swift
```diff
@@ -0,0 +1,167 @@
...
+
+    // MARK: - Background resolution
+
+    /// `.medium` can reach the glass appearance, so `.automatic` must hand back `nil` and let the
+    /// system make the per-detent call rather than pinning a color and killing the glass.
+    @Test
+    func automaticBackgroundDefersToSystemForMediumWhenLiquidGlassIsAvailable() {
+        let resolved = Warp.SheetBackground.automatic.resolvedColor(
+            token: Warp.Theme.token,
+            style: .medium,
```
luk-m-s, line 53: We call `Warp.Theme` global var directly, it would introduce shared state in isolated context :|

TBH, I don't know if it's actually affect tests results in any way because this test cover just liquid glass switch - just fyi.

Tests/WarpTests/Native/Sheet/SheetStyleModifierTests.swift
```diff
@@ -0,0 +1,167 @@
...
+        #expect(resolved == nil)
+    }
+
+    /// A full-screen sheet is opaque on every OS, so deferring to the system would only mean
+    /// taking Apple's color instead of the brand's. Visible in dark mode, where the system uses
+    /// `#1c1c1e` and `surfaceElevated100` is `#26262b`.
+    @Test
+    func automaticBackgroundUsesTheTokenForFullScreenEvenWithLiquidGlass() {
+        let token = Warp.Theme.token
+
```
luk-m-s, line 67: Just as above - maybe better use some hardcoded theme (like `.finn`) than fetching it from global var

Tests/WarpTests/Native/Sheet/SheetStyleModifierTests.swift
```diff
@@ -0,0 +1,167 @@
...
+            nativeGlassEffectsEnabled: true
+        )
+
+        #expect(components(resolved) == components(token.surfaceElevated100))
+    }
+
+    @Test(arguments: [Warp.SheetStyle.medium, .fullScreen])
+    func automaticBackgroundFallsBackToTokenBelowLiquidGlass(style: Warp.SheetStyle) {
+        let token = Warp.Theme.token
+
```
luk-m-s, line 81: As above

Tests/WarpTests/Native/Sheet/SheetStyleModifierTests.swift
```diff
@@ -0,0 +1,167 @@
...
+        #expect(components(resolved) == components(token.surfaceElevated100))
+    }
+
+    /// Offscreen snapshot hosts cannot rasterize native Liquid Glass, so disabling it must force
+    /// the opaque fallback even on an OS that supports glass.
+    @Test
+    func automaticBackgroundFallsBackToTokenWhenGlassEffectsAreDisabled() {
+        let token = Warp.Theme.token
+
```
luk-m-s, line 97: As above

Tests/WarpTests/Native/Sheet/SheetStyleModifierTests.swift
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    // MARK: - Brand coverage
+
+    /// `surfaceElevated100` differs per brand, so the fallback must follow the active theme
+    /// rather than resolving to a single hard-coded color.
+    @Test(arguments: Warp.Brand.allCases)
+    func opaqueFallbackFollowsActiveBrand(brand: Warp.Brand) {
+        let previousTheme = Warp.Theme
+        defer { Warp.Theme = previousTheme }
```
luk-m-s, line 155: Just FYI: Defer doesn't make sense here - in best case it simply do nothing because other tests setup it on their own, in worst case could create hard to debug random failings due to shared state of `Warp.Theme` in concurrent setting.

luk-m-s (changes requested):
Hardcoded accessibilityLabel for sure need fix, rest are commentaries :)

SampleApp/Components/Native/SheetDemoView.swift
MarcinGie: Fixed in 3b01b2e - and it turned out to be worse than a hardcoded string. `Warp.Icon.body` already applies `.accessibilityLabel(localization)` (`Icon.swift:311`), and `.close` resolves to `Warp.Strings.iconClose.localized` (`Icon.swift:421`), so the `.accessibilityLabel("Close")` on the Button was *overriding* the localized label with English for every brand and locale.

So rather than borrowing the label, I removed the override and let it travel up from the icon through the Button - which is how the sibling action button in the same toolbar already works.

Sources/Components/Native/Sheet/SheetStyleModifier.swift
MarcinGie: Extracted in 3b01b2e to `Sources/Helpers/LiquidGlass.swift` as `Warp.supportsLiquidGlass`.

One caveat worth flagging, because the new file looks like something the other iOS 26 branches should adopt and they can't: `if #available` / `guard #available` do two jobs at once - they test the OS *and* unlock the guarded API for the compiler. So `GlassControlContainer.makeGlassOrBlurEffect()` calling `UIGlassEffect(style:)` and `UIBarButtonItem+Style` setting `.prominent` have to keep the syntactic check; swapping in a `Bool` would not compile.

Of the ~15 iOS 26 sites in `Sources/`, the sheet is the only one that uses glass support as an *input to a decision* rather than a gate around a call, which is what makes the `Bool` form usable there at all. That's also the reason it's injected into `resolvedColor` instead of read inline - it's what lets the resolution be tested on any OS.

I've documented that limit in the file so the next person doesn't try to roll the others onto it. Happy to keep it inline in the modifier instead if a shared helper with one caller reads as over-abstraction to you.

Tests/WarpTests/Native/Sheet/SheetStyleModifierTests.swift
MarcinGie: Fixed in 3b01b2e. The suite now takes `Warp.Brand.finn.token` and never touches the global.

You were right that it didn't affect results, and the reason is that these cases assert a resolved background matches the token they were *handed* - which holds for any brand. So reading `Warp.Theme` bought nothing and only coupled them to state another suite could mutate. With the ambient dependency gone the suite drops `.serialized` and runs in parallel again, which is the actual win.

Tests/WarpTests/Native/Sheet/SheetStyleModifierTests.swift
MarcinGie: Removed in 3b01b2e, along with the `Warp.Theme = brand` write it was guarding.

Your instinct was right and the reason is stronger than shared state: the mutation had no effect on this test at all. `Brand.token` is a stateless per-brand provider - `Warp.swift:40` switches to a fresh `FinnTokenProvider()` and friends, each building colors from brand-specific constants with no global read - so `brand.token` fully determined the result and setting `Warp.Theme` was invisible to the assertion. The save/restore/defer was dead code protecting a write that didn't matter.

Test renamed to `opaqueFallbackFollowsBrand`, since "active brand" was describing a global it never actually read.

Tests/WarpTests/Native/Sheet/SheetStyleModifierTests.swift
MarcinGie: Covered by the reply on the first of these - the suite now uses `Warp.Brand.finn.token` throughout and no longer reads the global. Fixed in 3b01b2e.

Tests/WarpTests/Native/Sheet/SheetStyleModifierTests.swift
MarcinGie: Covered by the reply on the first of these - the suite now uses `Warp.Brand.finn.token` throughout and no longer reads the global. Fixed in 3b01b2e.

Tests/WarpTests/Native/Sheet/SheetStyleModifierTests.swift
MarcinGie: Covered by the reply on the first of these - the suite now uses `Warp.Brand.finn.token` throughout and no longer reads the global. Fixed in 3b01b2e.

SampleApp/Components/Native/SheetDemoView.swift
luk-m-s: 👍

Tests/WarpTests/Native/Sheet/SheetStyleModifierTests.swift
luk-m-s: 👌

Tests/WarpTests/Native/Sheet/SheetStyleModifierTests.swift
luk-m-s: 👍

Tests/WarpTests/Native/Sheet/SheetStyleModifierTests.swift
luk-m-s: 👍

Tests/WarpTests/Native/Sheet/SheetStyleModifierTests.swift
luk-m-s: 👍

Sources/Components/Native/Sheet/SheetStyleModifier.swift
luk-m-s: 🙇‍♂️

luk-m-s (approved):
Go, go, go!

PR #221, opened 2026-07-21
WARP-1409 Implement style modifiers for menu context menu

luk-m-s:
## What

Adds Warp design token support for native SwiftUI `Menu` and `.contextMenu`.

### `Warp.MenuButtonStyle`

Two cases — only what SwiftUI `ButtonRole` actually supports:
- `.default` — Warp icon token color
- `.destructive` — system red text via `ButtonRole.destructive`, Warp negative icon color

> Text color in menus is always OS-controlled (UIKit renders items). Only icon color and `ButtonRole` are affected.

### `Warp.MenuButton`

Leaf action item with optional Warp icon and style:

```swift
Warp.MenuButton("Share", icon: .share) { share() }
Warp.MenuButton("Delete", icon: .bin, style: .destructive) { delete() }
```

### `Warp.Menu`

Submenu container with `@ViewBuilder` content — composes with `Warp.MenuButton` or nests into other `Warp.Menu`:

```swift
Warp.Menu("Actions") {
    Warp.MenuButton("Share", icon: .share) { share() }
    Warp.Menu("More", icon: .dots) {
        Warp.MenuButton("Bookmark", icon: .bookmark) { bookmark() }
    }
    Warp.MenuButton("Delete", icon: .bin, style: .destructive) { delete() }
}
```

Both work in `.contextMenu` closures.

### `View.warpMenuButton(style:)`

Modifier for styling raw `SwiftUI.Menu` trigger labels with Warp tint and font tokens.

## Snapshot tests

`MenuSnapshotTests` covers all 6 brands:
- `snapshotMenuButtonStyles` — default and destructive trigger styles
- `snapshotMenuItems` — text-only, with Warp icon, destructive variants


Sources/Components/Native/Menu/Menu+Style.swift
```diff
@@ -0,0 +1,118 @@
...
+    /// ```
+    ///
+    /// - Parameter style: The visual style to apply. Defaults to `.default`.
+    /// - Returns: A view with Warp menu button styling applied.
+    public func warpMenuButton(style: Warp.MenuButtonStyle = .default) -> some View {
+        modifier(WarpMenuButtonModifier(style: style))
+    }
+}
+
+private struct WarpMenuButtonModifier: ViewModifier {
```
MarcinGie, line 26: The PR description mentions this is "consistent with the pattern established by the navigation bar and tab bar native component PRs" — the API shape definitely is, nice work there. One thing though: `ToolbarItem+Style.swift` (the SwiftUI-side nav bar equivalent) resolves colors through `@Environment(\.warpTheme)` and a `colorProvider` (see lines 102–124 in that file), while `WarpMenuButtonModifier` here uses static `Warp.Color.token` accessors directly.

In practice this means `.warpMenuButton()` won't pick up per-view theme overrides set via `.warpTheme()`. It works fine when the global theme is the only one in play, but it'd be good to align with the `ToolbarItem` approach so the whole native component family behaves the same way.

Sources/Components/Native/Menu/Menu+Style.swift
```diff
@@ -0,0 +1,118 @@
...
+    /// Applies Warp design tokens (color, font) to a `Menu` or `ContextMenu` trigger.
+    ///
+    /// Use this modifier on any `Menu` label view to apply consistent Warp brand
+    /// colors and typography.
+    ///
+    /// **Usage:**
+    ///
+    /// ```swift
+    /// Menu("Options") {
+    ///     Warp.MenuItem("Share", icon: .shareIOS) { share() }
```
MarcinGie: Small consistency thing — the doc comments here use `.shareIOS` and `.delete`:
`Warp.MenuItem("Share", icon: .shareIOS) { share() }`
`Warp.MenuItem("Delete", icon: .delete, role: .destructive) { delete() }`

But the PR description shows `.share` and `.heartRate` for the same actions. It'd be nice to align these so someone copy-pasting from either source gets working, sensible examples. The doc comment versions (`.shareIOS` / `.delete`) are arguably the better choice since the icon names match the action semantically.

Tests/WarpSnapshotTests/Components/Native/Menu/MenuSnapshotTests.swift
```diff
@@ -0,0 +1,85 @@
...
+                    Warp.MenuItem("Bookmark", icon: .bookmark) {}
+                    Warp.MenuItem("Edit", icon: .edit) {}
+                    Warp.MenuItem("Copy", icon: .copy) {}
+                }
+            }
+
+            GroupBox("Destructive") {
+                VStack(alignment: .leading, spacing: 4) {
+                    Warp.MenuItem("Delete", role: .destructive) {}
+                    Warp.MenuItem("Remove", icon: .heartRate, role: .destructive) {}
```
MarcinGie: This pairs a heart rate icon with a "Remove" action — it works as a test case but might read oddly to someone looking at snapshot baselines later and wondering why a medical icon is on a destructive button. If there's a more semantically fitting icon available (like `.trashcan` or `.close`), it'd make the test snapshot more self-documenting. Totally optional — the test logic is fine either way.

Tests/WarpSnapshotTests/Components/Native/Menu/MenuSnapshotTests.swift
```diff
@@ -0,0 +1,85 @@
...
+        switch self {
+        case .default: return "Default"
+        case .primary: return "Primary"
+        case .destructive: return "Destructive"
+        }
+    }
+}
+
+extension Warp.MenuButtonStyle: Identifiable {
+    public var id: String { label }
```
MarcinGie, line 83: This adds `public Identifiable` conformance to `Warp.MenuButtonStyle`. Since this lives in the test target, the `public` access level doesn't cause real issues — but it's a protocol conformance on a library type, which Swift considers module-wide. If `Identifiable` conformance is genuinely useful, it might belong in the main source alongside the enum itself. If it's only for this test's `ForEach`, dropping the access level or using a wrapper type would be a bit cleaner. Not a blocker at all, just flagging it.

Sources/Components/Native/Menu/Menu+Style.swift
```diff
@@ -0,0 +1,118 @@
...
+    ///     Warp.MenuItem("Delete", icon: .delete, role: .destructive) { delete() }
+    /// }
+    /// ```
+    ///
+    /// Text-only (no icon):
+    ///
+    /// ```swift
+    /// Warp.MenuItem("Cancel", role: .cancel) { }
+    /// ```
+    public struct MenuItem: View {
```
MarcinGie: Same theme observation as the modifier - `MenuItem` doesn't read `@Environment(\.warpTheme)`. It delegates to `Warp.IconView` for the icon (line 110), which may internally be theme-aware, but the `MenuItem` itself doesn't have access to the color provider. Right now this doesn't visibly matter since `MenuItem` doesn't apply custom colors - the `ButtonRole` handles destructive styling via the system. But if `MenuItem` ever needs to apply Warp token colors (e.g. for a future `.subtle` style or custom text color), the environment hookup would need to be added then. Just worth being aware of - no action needed today.

MarcinGie (approved):
Nice, clean PR — the API shape (`MenuButtonStyle` enum + `.warpMenuButton()` modifier + `Warp.MenuItem` view) follows the pattern from the nav bar/tab bar PRs well, and the snapshot coverage across all 6 brands is solid. The demo view is interactive and exercises both `Menu` and `.contextMenu` which is great for manual testing.

The main thing I'd flag is the theming approach: `ToolbarItem+Style.swift` reads `@Environment(\.warpTheme)` and resolves colors through `colorProvider`, while this PR uses static `Warp.Color.token` accessors throughout. It works fine for the global theme case, but means `.warpMenuButton()` won't respond to per-view `.warpTheme()` overrides. Worth aligning so the native component family behaves consistently.

The icon examples in the PR description (`.share`, `.heartRate`) differ from the doc comments in the code (`.shareIOS`, `.delete`) — would be good to align those so copy-pasting from either source gives sensible results.

Other than that, this looks good to go 👍  The smaller items (test-only `Identifiable` conformance, `.heartRate` for "Remove" in snapshots) are just polish — nothing blocking.

Sources/Components/Native/Menu/Menu+Style.swift
luk-m-s: Fixed

Sources/Components/Native/Menu/Menu+Style.swift
luk-m-s: I've changed this

Tests/WarpSnapshotTests/Components/Native/Menu/MenuSnapshotTests.swift
luk-m-s: 👍

Sources/Components/Native/Menu/Menu+Style.swift
luk-m-s: Fixed.

PR #223, opened 2026-07-27
WARP-1411 Make Warp tokens available via `theme` environment property

luk-m-s:
### What

Refactors the token system to eliminate global state usage and align with the environment-based theme injection pattern already established by `@Environment(\.warpTheme)`.

### Changes

- Made `token: TokenProvider` and `uiToken: UITokenProvider` properties `public` with environment-based usage docs
- `Warp.Token` and `Warp.UIToken` static vars now delegate to `Theme.token` / `Theme.uiToken` — removing duplicated switch statements
- **`WarpNavigationBarButtonModifier`** — replaces static `Warp.Color.token` usage with `@Environment(\.warpTheme)`, resolving both `token` and `colors` from the environment (addresses known feedback from previous PR reviews)
- **Public API change** — `Color` default parameters in `BorderOptions` and `View+` helpers changed from `Color` to `Color? = nil`; default is now resolved internally from the theme. Existing call sites passing a non-optional `Color` still compile, but the contract changed from `"defaults to Warp.Token.border"` to `"defaults to nil, resolved from theme"` — worth verifying in client repos
- Removed unused `colorProvider` computed property from `ConfirmationDialogViewModifier`

### Testing

- `MetricsFactorySmokeTests` updated to supply `theme:` in `UIMetricsFactory` init


Sources/Helpers/BorderOptions.swift
```diff
@@ -18,76 +18,78 @@ import SwiftUI
...
-    ///
     /// - Parameters:
-    ///   - color: The color of the border. Defaults to `Warp.Color.token.border`.
+    ///   - color: The color of the border. Defaults to `token.border` from the environment theme.
     ///   - width: The thickness of the border. Defaults to `1`.
     ///   - sides: A `BorderOptions` option set specifying the sides where the border should be applied.
     /// - Returns: A view with borders on the specified sides.
-    public func border(_ color: Color = Warp.Token.border, width: CGFloat = 1, sides: BorderOptions) -> some View {
-        overlay(
+    public func border(_ color: Color? = nil, width: CGFloat = 1, sides: BorderOptions) -> some View {
```
MarcinGie, line 50: The type changed from `Color` to `Color?`. Any call site passing a non-optional `Color` still compiles, but any code that relied on the return type being inferred from the non-optional parameter (unlikely but possible) could break. More importantly, this is a public API change - consumers using this library who call `.border(sides: .top)` will still get the right default, but the contract changed from "defaults to Warp.Token.border" to "defaults to nil, resolved internally." Worth a note in the PR description. Probably also a check in the client repo if that breaks the current usage.

Sources/Helpers/View+.swift
```diff
@@ -22,46 +22,65 @@ extension View {
...
-    ///
     /// - Parameters:
-    ///   - color: The color to be used for the background. Defaults to `Warp.Token.surfaceElevated200`.
+    ///   - color: The color to be used for the background. Defaults to `token.surfaceElevated200` from the environment theme.
     ///   - cornerRadius: The radius of the rounded corners. Defaults to `Warp.Spacing.spacing200`.
     ///   - isPressed: A value indicating whether the view is currently in a pressed state and should update its shadow accordingly. Defaults to `false`.
     /// - Returns: A view with the applied card-like background and shadow.
     public func withCardBackground(
-        _ color: Color = Warp.Token.surfaceElevated200,
+        _ color: Color? = nil,
```
MarcinGie, line 34: Same pattern - `Color` default changed to `Color? = nil`. Same implications as above.

Sources/Helpers/View+.swift
```diff
@@ -22,46 +22,65 @@ extension View {
...
-    /// the safe area to extend the view fully.
-    ///
     /// - Parameters:
-    ///   - color: The background color for the bottom sheet. Defaults to `Warp.Token.surfaceElevated100`.
+    ///   - color: The background color for the bottom sheet. Defaults to `token.surfaceElevated100` from the environment theme.
     ///   - cornerRadius: The radius for the rounded top corners. Defaults to `Warp.Spacing.spacing200`.
     /// - Returns: A view with the applied bottom sheet background, rounded top corners, and shadow.
     public func withBottomSheetBackground(
-        _ color: Color = Warp.Token.surfaceElevated100,
+        _ color: Color? = nil,
```
MarcinGie, line 48: Same pattern - `Color` default changed to `Color? = nil`. Same implications as above.

Sources/Components/ConfirmationDialog/ConfirmationDialogViewModifier.swift
MarcinGie: Untouched nitpic: `colorProvider` isn't used anywhere in this file - only `token` is (lines 36–37). If `colorProvider` was there for future use, that's fine, but if it's leftover from before the refactor, it could be cleaned up.

Sources/Components/Native/NavigationBar/Button+Style.swift
```diff
@@ -1,5 +1,29 @@
 import SwiftUI
 
+@available(iOS 26.0, *)
```
MarcinGie, line 3: This is one of the more impactful changes - it was the exact issue flagged in previous PR reviews (using static `Warp.Color.token` instead of environment). The new `WarpNavigationBarButtonModifier` properly reads `@Environment(\.warpTheme)` and resolves both `token` and `colors`. This is the right fix. Worth calling out in the PR description since it addresses known feedback.

MarcinGie:
The PR description says `MetricsFactorySmokeTests` was updated, which is accurate - but there are no new tests for the refactored modifiers (`BorderViewModifier`, `CardBackgroundModifier`, `BottomSheetBackgroundModifier`, `WarpNavigationBarButtonModifier`). Since these are behavioral refactors (same output, different token resolution path), snapshot tests would catch regressions if they exist. If there are existing snapshot tests covering these components, that's sufficient - just worth confirming.

MarcinGie (approved):
This is a high-quality refactor that addresses the exact global-state issue flagged in multiple previous PRs. The pattern is consistent - extract a `ViewModifier`, read `@Environment(\.warpTheme)`, resolve tokens through the theme. The `Warp.Token`/`Warp.UIToken` delegation cleanup is clean. Main feedback: update the PR description to reflect the full scope (it's doing more good work than it claims), and consider removing the unused `colorProvider` in `ConfirmationDialogViewModifier`.

Sources/Helpers/BorderOptions.swift
luk-m-s: > but any code that relied on the return type being inferred from the non-optional parameter (unlikely but possible) could break.

Type changes to make contract more flexible and allows to completely rid dependance on global mutable module variable `Warp.Token`. I don't see any possible scenario when passing `nil` here instead of concrete variable could broke output type inference - especially when compiler sees implementation code (this module isn't shipped as binary artifact).

Sources/Helpers/View+.swift
luk-m-s: Same as above

Sources/Helpers/View+.swift
luk-m-s: See above reply

Sources/Components/ConfirmationDialog/ConfirmationDialogViewModifier.swift
luk-m-s: Removed.

Sources/Components/Native/NavigationBar/Button+Style.swift
luk-m-s: Yep but there is still `Warp.Typography` leftover which need to be refactored

luk-m-s: PR updated

PR #226, opened 2026-08-10
WARP-1417 Refactor font token access and snapshot tests to use SwiftUI environment injection

luk-m-s:
## What

Fixes Button race condition during theme injection setup in snapshot tests.

## Changes

- Removed `Font` from public `Brand` enum exposure (was leaking internal concerns)
- Snapshot tests now set global `Warp` theme AND inject via `.environment(\.warpTheme, ...)` to prevent theme mismatch during UIKit rendering
- Fixed font registration logic that caused race conditions when tests ran with concurrent snapshot recording

## Testing

- All 30+ snapshot tests passing with stable theme state
- UIKit-backed components render correctly with synchronized theme setup


Sources/Components/AlertDialog/AlertDialogViewModifier.swift
```diff
@@ -16,6 +16,8 @@ extension Warp {
             theme.colors
         }
 
+        private var fonts: [Warp.Font] { theme.fonts }
```
MarcinGie: I don't think anything reads `fonts` - line 35 still calls `.font(Warp.Typography.title4.font)`, which goes through the global theme. Swift doesn't warn on an unused private computed property, so this compiles clean while changing nothing.

  Worth saying why it looks plausible but isn't: `Warp.Font` is the font-*file* enum (`.finnMedium`, `.toriBold`) used for registration, not a typography token, so a `[Warp.Font]` array doesn't really have a meaning inside a view body. This file also never referenced `Warp.Font` before the PR, so there's no global usage being removed here either.

  Same property landed in four other files, all unused with the `Warp.Typography` call still in place:

  - `Sources/Components/ConfirmationDialog/ConfirmationDialogViewModifier.swift:18` (uses `Warp.Typography.title4` at `:32`)
  - `Sources/Components/Native/NavigationBar/Button+Style.swift:9` (uses `Warp.Typography` at `:18` and `:23`)
  - `Sources/Components/TextArea/TextArea.swift:41` (uses `Warp.Typography.body` at `:153`)
  - `Sources/Components/TextField/TextField.swift:44` (uses `Warp.Typography.body` at `:218`)

  I'd suggest just deleting all five - the real change would be making `Warp.Typography` brand-aware, which is a bigger piece.

Sources/Typography/Font.swift
```diff
@@ -5,7 +5,7 @@ extension Warp {
     ///
     /// Each case corresponds to a specific font used by different brands (e.g., `finn`, `tori`, `dba`, `blocket`), and each font has
     /// associated file names and font names that are used for registration and application in the design system.
-    enum Font: CaseIterable, Sendable {
+    public enum Font: CaseIterable, Sendable {
```
MarcinGie: Non-blocking, but: making this `public` while `fileName` (`:39`), `name` (`:60`) and `fonts(for:)` (`:82`) stay internal means external consumers get an enum they can enumerate but can't inspect. The only thing reachable is `Brand.fonts`, which hands back opaque cases.

  Since `Warp.Typography.registerFonts(for:)` is already public and covers the use case, is the public exposure actually needed here?

  Semi-related: `ENVIRONMENT_MIGRATION.md:167` tells consumers to use `Warp.Font.fonts(for:)`, which is internal. Pre-existing, but this PR makes the surrounding API public without fixing it, so it might be a good moment.

Sources/Warp.swift
```diff
@@ -100,8 +100,14 @@ public enum Warp {
...
 
-        /// Fonts for this brand
-        var fonts: [Warp.Font] {
+        /// Fonts associated with this brand, used for font registration.
+        ///
+        /// Prefer reading this through `@Environment(\.warpTheme)` in SwiftUI views:
+        /// ```swift
+        /// @Environment(\.warpTheme) private var theme
+        /// private var fonts: [Warp.Font] { theme.fonts }
+        /// ```
```
MarcinGie: I think this example is what produced the five unused `fonts` properties. `Brand.fonts` is a font-registration list, so there's no real reason to read it from a view's environment, and doing so looks like token access when it isn't.

  Suggest keeping the first line ("Fonts associated with this brand, used for font registration") and dropping the `@Environment` snippet - or swapping it for the actual call site, `Warp.Typography.registerFonts(for: theme)`.

Tests/WarpSnapshotTests/Components/TextSnapshotTests.swift
```diff
@@ -28,6 +25,6 @@ struct TextSnapshotTests {
         // Set width to match iPhone 13 size
         .frame(width: ViewImageConfig.iPhone13.size!.width)
 
-        assertSnapshot(of: textViewsInColumn, as: .warpImage(compressionQuality: .high), named: snapshotName)
+        assertSnapshot(of: textViewsInColumn.warpTheme(brand), as: .warpImage(compressionQuality: .high), named: snapshotName)
```
MarcinGie, line 29: This is the one I'd most like to sort before merge: `.warpTheme(brand)` on its own doesn't drive typography.

  `Warp.Typography.font` / `.uiFont` switch on the global `Warp.Theme` (`Sources/Typography/Typography.swift:131,149,167,185`) and never read `\.warpTheme`. And font *registration* only fires from the `Warp.Theme` `didSet` (`Sources/Warp.swift:128-133`), so with the assignment removed no brand fonts get registered in the snapshot process at all - text falls back
  to the system font.

  Repro on one simulator (iPhone 16 / iOS 26.0), one suite at a time:

  - `main`: `TextSnapshotTests` passes 6/6
  - this branch: fails 6/6
  - this branch with `Warp.Theme = brand` restored *and* `.warpTheme(brand)` kept: passes 6/6

  The diffs are metric rather than colour (`Newly-taken snapshot@(390.0, 185.33) does not match reference@(390.0, 204.0)`), and `PageIndicatorSnapshotTests` (no text) passes on both, which is consistent with fonts being the cause.

  Same applies to every snapshot file in this PR that dropped `Warp.Theme = brand`. Either keep the assignment until typography is environment-aware, or make `Warp.Typography` take a `Warp.Brand` - the latter being the actual legacy this ticket names.

MarcinGie (commented):
Thanks for pushing on the token migration - the `brand.token.x` conversions in the Checkbox/Radio/Icon tests, the `ColorFactorySmokeTests` cleanup, and `theme.fonts` in `Typography+Register` are all genuine improvements and I'd like to keep them.

  Two things I'd like to sort before merge though, details inline.

  ### 1. The snapshot migration turns brand fonts off

  Swapping `Warp.Theme = brand` for `.warpTheme(brand)` doesn't carry typography across: `Warp.Typography.font` / `.uiFont` switch on the global `Warp.Theme` (`Sources/Typography/Typography.swift:131,149,167,185`) and never read `\.warpTheme`. On top of that, font *registration* only fires from the `Warp.Theme` `didSet` (`Sources/Warp.swift:128-133`), so with
  the assignment gone no brand fonts get registered in the snapshot process at all and text falls back to the system font.

  I ran it a few ways on the same simulator (iPhone 16 / iOS 26.0), one suite at a time:

  | | result |
  |---|---|
  | `main` | `TextSnapshotTests` passes 6/6 |
  | this branch | fails 6/6 |
  | this branch, `Warp.Theme = brand` restored *and* `.warpTheme(brand)` kept | passes 6/6 |

  The diffs are metric changes rather than colour (`Newly-taken snapshot@(390.0, 185.33) does not match reference@(390.0, 204.0)`), and `PageIndicatorSnapshotTests` (no text in it) passes on both, which fits.

  ### 2. The component refactor doesn't currently do anything

  The five new `private var fonts: [Warp.Font] { theme.fonts }` properties are unused - those files still resolve fonts via `Warp.Typography`, and none of them referenced `Warp.Font` before this PR, so there's no global usage being removed. Swift doesn't warn on an unused private computed property, so it all compiles clean.

  ### On the description

  Worth double-checking how the snapshot suite was verified, since CI doesn't run it - `run-tests.yml` only runs `-scheme WarpTests`, so the 33 `WarpSnapshotTests` suites never execute on a PR. That's a separate problem and not yours to fix here; I'll open a ticket for gating CI on the snapshot scheme rather than piling it onto this PR.

  ### Two ways forward, your call

  - **Scope down:** drop the five properties, restore `Warp.Theme = brand` in the snapshot tests, keep the test-side token cleanups. Mergeable today.
  - **Do the full refactor:** make `Warp.Typography` accept a `Warp.Brand` and have components resolve fonts through it. That's what WARP-1417 actually points at and what would make the injection meaningful - but it needs reference re-recording.

  I'd lean towards the first here plus a follow-up ticket for the second, but happy either way.

Sources/Components/AlertDialog/AlertDialogViewModifier.swift
luk-m-s: Thanks, I removed all of them for now 👍

Sources/Warp.swift
luk-m-s: Fixed

Tests/WarpSnapshotTests/Components/TextSnapshotTests.swift
luk-m-s: Yeah, font registration is one main culprit with SwiftUI environment injection because API doesn't provide you feedback if font is already registered.

In production app this is not really a problem because you must call `Warp.Theme` set at least once (for UIKit components) but it becomes problem in isolated snapshot/unit tests cases

I restore setting `Warp.Theme` property for affected snapshot tests

luk-m-s:
I also choose first approach. For now I don't see any practical way to remove `registerFonts` method because used APIs is very old. SwiftUI environment injection doesn't help there too.

Normally this shouldn't be a problem because in prod app we don't switching brand/themes dynamically but only once per launch - so you don't have to remember to calling `registerFonts` manually because setting global var `Warp.Theme` do it for you (and you cannot eliminate global var either because it has its use in UIKit components).

I scope down this ticket and need to think if there is any way to workaround font registering issue in future.

Sources/Typography/Font.swift
luk-m-s: Done.

MarcinGie (approved):
Re-reviewed at `e1057c3`. Both blockers are resolved, and I verified the snapshot one by running it rather than reading it, so this is good to go from my side.

  **Blocker 1 (snapshot fonts) - fixed and verified.** iPhone 16 / iOS 26.0, one suite at a time:

  | | `e02dcef` (before) | `e1057c3` (now) |
  |---|---|---|
  | `TextSnapshotTests` | 6/6 fail | 6/6 pass |
  | `ButtonSnapshotTests` | - | 48/48 pass |

  **Blocker 2 (unused `fonts` properties) - fixed.** All five removed, `Warp.Font` back to internal, `Brand.fonts` doc comment corrected.

  **An unexpected win worth recording.** The `Warp.Token` to `brand.token.x` conversions fix a real race, not just style. On `main`, `ButtonSnapshotTests` fails 28 of its 48 cases *in isolation*: the parameterized cases run in parallel, each one assigns the global
  `Warp.Theme`, and the view bodies then read `Warp.Token` from whichever brand won. On this branch the same suite passes 48/48. Across the whole scheme in one process: `main` 364 issues in 9 suites, this branch 157 issues in 4. That is the strongest argument for the PR as it now stands, and it is nowhere in the description.

  **The remaining red is not yours.** Every suite still failing on the branch fails identically on `main`: `SnackbarSnapshotTests` (144 of 216 on both, run in isolation), `IconSnapshotTests.snapshotRegularIcons()`, and `MenuSnapshotTests` /
  `WarpTabBarBadgeStylingSnapshotTests` 6/6 each, both untouched here and inherited from #221 and #227. I will fold those into the snapshot-CI ticket rather than this PR.

  **One thing to fix before merge (non-blocking):** the description is now stale and will land as the squash commit message. It still says the five components read fonts from `@Environment(\.warpTheme)` (reverted), that `fonts` was added to `Brand` (it already existed on `main`; the only remaining `Warp.swift` change is the doc comment), and that the snapshot tests were migrated to injection (they now set the global *and* inject). Worth rewriting to match the scoped-down reality you describe in the thread, and the Button race above is the headline it deserves.

PR #227, opened 2026-08-14
Fix solid-fill presence badge on tab bar

teddyfahi:
# What
  `warpSetBadgeDot()` set `badgeValue = "•"`, which shows a visible dot inside the badge (a hollow
  doughnut on iOS 26+). Figma spec is a solid fill.

  Fix: set `badgeValue = " "` and rename to `warpSetPresenceBadge()`. Old name kept as
  `@available(*, deprecated, renamed:)` alias so existing consumers don't break.
  
# Why?
Inconsitency in how the badge dot is rendered. Design approved: [FINN Slack thread, 
  2026-08-07](https://sch-chat.slack.com/archives/C0B2C977Z5J/p1785931019049309).


SampleApp/Components/Native/TabBar.swift
```diff
@@ -26,7 +26,7 @@ struct WarpTabBarControllerWrapper: UIViewControllerRepresentable {
             guard let item else { continue }
             switch badges[index] {
             case .count(let value): item.warpSetBadge(count: value)
-            case .dot:              item.warpSetBadgeDot()
+            case .dot:              item.warpSetPresenceBadge()
```
luk-m-s, line 29: I'm not sure if leaving old `dot` case could be a little misleading now but it's only semantic

luk-m-s (approved):
🚀

PR #228, opened 2026-08-17
WARP-1423 Migrate scrollable SegmentedControl from NMP to Warp (as part of native components)

luk-m-s:
## Why

Scrollable segmented control was previously living in NMP. Moving it to Warp makes it available to all consumers and aligns with the design system's native components roadmap.

## What

Adds a scrollable segmented control component to the Warp library, migrated from NMP as part of the native components initiative.

### New public API

- `GlassSegmentedControl` — UIKit component with delegate-based selection callback
- `Warp.SegmentedControl` — SwiftUI `UIViewRepresentable` wrapper with `Binding<String?>` for selected identifier

## Testing

- Demo screen added to `WarpDemo` under `SegmentedControlDemoView`
- Manual testing: tab selection, scrolling to selected segment, iOS 26 glass vs pre-iOS 26 tab strip rendering


Sources/Components/Native/SegmentedControl/ScrollableTabview.swift
```diff
@@ -0,0 +1,212 @@
+import UIKit
```
MarcinGie, line 1: Nitpic: filename case: should be `ScrollableTabView`

Sources/Components/Native/SegmentedControl/ScrollableTabview.swift
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    func configure(items: [GlassSegmentedControl.Item], selectedIdentifier: String?) {
+        self.items = items
+        var snap = NSDiffableDataSourceSnapshot<Int, GlassSegmentedControl.Item>()
+        snap.appendSections([0])
+        snap.appendItems(items, toSection: 0)
+        dataSource.apply(snap)
+
+        if let id = selectedIdentifier,
```
MarcinGie: I think the deselect path from the FinniversKit original got lost here. The original passed an **optional** `IndexPath`, so `nil` cleared the selection:

  ```swift
  let selectedIndex = viewModel.items.firstIndex(where: { $0.identifier == viewModel.selectedIdentifier })
  collectionView.selectItem(at: indexPath(for: selectedIndex), animated: false, scrollPosition: .centeredHorizontally)
  ```

  With the `if let`, when `selectedIdentifier` is `nil` - or points at an id that's no longer in `items` - the previously selected cell stays highlighted.

  Both states are reachable straight from the demo screen this PR adds: the Picker's `Text("none").tag(String?.none)` row, and `onDelete` on the currently-selected item.

  Worth noting the iOS 26 branch *does* clear on the same input (`removeAllSegments()` leaves `selectedSegmentIndex` at `.noSegment`), so the two paths disagree today. Restoring the optional-`IndexPath` call fixes the regression and the divergence in one go. 🙂

Sources/Components/Native/SegmentedControl/ScrollableTabview.swift
```diff
@@ -0,0 +1,212 @@
...
+        collectionView.scrollToItem(at: indexPath, at: .centeredHorizontally, animated: true)
+    }
+}
+
+// MARK: - TabCell
+
+private final class TabCell: UICollectionViewCell {
+    static let reuseID = "TabCell"
+
+    private static let font = Warp.Typography.captionStrong.uiFont
```
MarcinGie, line 128: Small line, but I think this one bites: `Warp.Typography.captionStrong.uiFont` resolves against the global `Warp.Theme` (`Typography.swift:165-181`), and a `static let` initialises exactly once per process — so whichever brand happens to render the first tab wins for the lifetime of the app. It's reproducible in the SampleApp today, which switches brands at
  runtime (`ContentView.swift:101`). `cellHeight` on line 137 is computed from it and inherits the same staleness.

  FinniversKit stored the *style* rather than the resolved font, which is why it didn't hit this:

  ```swift
  private static let labelStyle = Warp.Typography.captionStrong   // resolve .uiFont per access
  ```

  That's the minimal fix. Resolving from the injected `warpTheme` (the theming point in my summary) would be the proper one — either is fine by me.

Sources/Components/Native/SegmentedControl/SegmentedControl.swift
```diff
@@ -0,0 +1,97 @@
...
+            let control = GlassSegmentedControl()
+            control.translatesAutoresizingMaskIntoConstraints = true
+            control.delegate = context.coordinator
+            return control
+        }
+
+        public func sizeThatFits(_ proposal: ProposedViewSize, uiView: GlassSegmentedControl, context: Context) -> CGSize? {
+            let height = uiView.intrinsicContentSize.height
+            guard height != UIView.noIntrinsicMetric, height > 0 else { return nil }
+            return CGSize(width: proposal.width ?? 0, height: height)
```
MarcinGie: Does `proposal.width ?? 0` collapse the control to zero width in any context that doesn't propose one - `.fixedSize()`, a horizontal `ScrollView`, some `HStack` sizing passes?

  The `#Preview` below uses `.fixedSize(horizontal: false, vertical: true)` so a width proposal survives and it looks fine there, but a caller doing a plain `.fixedSize()` would get nothing. `uiView.systemLayoutSizeFitting(...)` could give a usable ideal width for the `nil` case.

Sources/Components/Native/SegmentedControl/GlassSegmentedControl.swift
```diff
@@ -0,0 +1,235 @@
...
+        guard items.count > 1 else { return }
+        layoutIfNeeded()
+
+        let totalWidth = segmentedControl.bounds.width
+        let visibleWidth = scrollView.bounds.width
+        guard totalWidth > visibleWidth else { return }
+
+        let segmentCenter = totalWidth * (CGFloat(index) + 0.5) / CGFloat(items.count)
+        let maxOffset = totalWidth - visibleWidth
+        let targetOffset = max(0, min(segmentCenter - visibleWidth / 2, maxOffset))
```
MarcinGie, line 216: Not a regression - this came over from the Drive version as-is - but flagging while we're here: the centring maths assumes uniform segment widths, whereas line 65 sets `apportionsSegmentWidthsByContent = true`, so widths track title length. The error compounds toward the end of the strip; the `#Preview` data in `SegmentedControl.swift` (`"First"` … `"Fifth tab
  with long title"`) shows it off nicely.

  Summing the actual per-segment widths (or reading the sorted subview frames) would give a correct centre. Totally fine as a follow-up - the migration just felt like a natural moment to mention it.

Sources/Components/Native/SegmentedControl/GlassSegmentedControl.swift
```diff
@@ -0,0 +1,235 @@
...
+        self.items = items
+
+        if usesGlassSegments {
+            configureSegmentedControl(selectedIdentifier: selectedIdentifier)
+        } else {
+            configureScrollableTabView(selectedIdentifier: selectedIdentifier)
+        }
+    }
+
+    public func updateShadow(using scrollView: UIScrollView, scrollViewParentView: UIView?) {
```
MarcinGie, line 117: `updateShadow(using:scrollViewParentView:)` doesn't look reachable from `Warp.SegmentedControl`, so `shadowRadius` stays at its initial `0` and the whole shadow apparatus in `ScrollableTabView` (`:30-34` setup, `:51-56` in `layoutSubviews`, `:75-85`) never does anything.

  If the public UIKit class is meant to be driven directly by NMP then this is live API and you can ignore me - it ties into the public-surface question in my summary. If `Warp.SegmentedControl` is the only intended entry point, it might be worth dropping.

Sources/Components/Native/SegmentedControl/GlassSegmentedControl.swift
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    private func setupGlassSegments() {
+        addSubview(glassContainer)
+        scrollView.addSubview(segmentedControl)
+
+        NSLayoutConstraint.activate([
+            glassContainer.topAnchor.constraint(equalTo: topAnchor),
+            glassContainer.leadingAnchor.constraint(equalTo: safeAreaLayoutGuide.leadingAnchor, constant: Warp.Spacing.spacing200),
+            glassContainer.trailingAnchor.constraint(equalTo: safeAreaLayoutGuide.trailingAnchor, constant: -Warp.Spacing.spacing200),
```
MarcinGie: Pinning to `safeAreaLayoutGuide ± Warp.Spacing.spacing200` bakes the caller's outer margin into the component. That's exactly right for the Drive fullscreen gallery it came from, but as a general Warp component it takes horizontal placement away from the consumer - they can't put it in a container that already has its own margins without doubling up.

  The pre-26 path doesn't do the equivalent either: it pins the tab view edge-to-edge (`:171-174`) and insets *inside* the collection section instead (`ScrollableTabview.swift:98-101`). Same visual result at the default, but the two diverge as soon as someone wants to change it. wdyt?

MarcinGie (changes requested):
Really nice migration - the Drive/FinniversKit code came across faithfully, and dropping the `ScrollableTabViewModel` and FinniversKit `Label` dependencies in favour of plain UIKit is a lovely simplification. I built the branch locally (`xcodebuild -scheme warp-ios -destination 'generic/platform=iOS Simulator'`) and it succeeds with no new warnings from these files. 👍

  I've left a few inline comments. Two I'd like to sort before merge (the deselect path and the static font); the rest are questions and suggestions - very happy to be talked out of any of them.

  ### Theming doesn't go through `@Environment(\.warpTheme)`

  All three source files read the global `Warp.UIToken` / `Warp.Typography` and capture the results inside `lazy var` initialisers. That's a bit at odds with `ENVIRONMENT_MIGRATION.md` and with `356f062 Use SwiftUI theme environment injection`, which the rest of the components have moved to.

  Two concrete consequences beyond convention: `.warpTheme(.blocket)` on a subtree has no effect on this control, and because the colours are captured once at lazy-init it never restyles even when the *global* brand changes — the `updateUIView` early-return would skip reconfiguration anyway. The `static let font` I flagged on `ScrollableTabview.swift:133` is one
  instance of this.

  `Warp.SegmentedControl` is a SwiftUI view, so it can read `@Environment(\.warpTheme)` and pass a `UITokenProvider` down into `GlassSegmentedControl.init` / `configure`. Multi-brand is sort of the core promise of Warp so I'd rather we didn't land a component that opts out — but if you'd prefer to ship the migration now and follow up, a scoped ticket works fine
  for me. Your call.

  ### Public API surface - a question, not a defect

  The description lists `GlassSegmentedControl` as intentional public API, so I'll ask rather than assert: it and `GlassSegmentedControlDelegate` land at module scope, while every other user-facing Warp type lives under `extension Warp`. The Drive-flavoured name ends up in every consumer's global namespace and gets expensive to rename after release - `Warp.GlassSegmentedControl` would match the rest of the library.

  Is the UIKit-facing API a hard requirement (i.e. NMP will call it directly), or is `Warp.SegmentedControl` the only intended entry point?

  ### Overlap with `Warp.Tabs`

  `Sources/Components/Tabs/Tabs.swift` is already a scrollable tab strip with an underline indicator, subtle→selected text colouring and centre-on-select - quite close to what the pre-iOS-26 `ScrollableTabView` reimplements in UIKit, but with different tokens (`Tabs` uses `token.textLink` + `token.border`; this uses `UIToken.text` + `UIToken.backgroundPrimary`).

  Is the visual divergence intentional per Figma? And is there a reason the fallback couldn't host `Warp.Tabs`, keeping `UIViewRepresentable` for the iOS 26 glass path only? Either way it'd be good for the docs to say when to reach for which.

  ### Tests

  I see the description says manual testing only. The other two native components have automated coverage (`Tests/WarpTests/Native/Sheet/SheetStyleModifierTests.swift`, `Tests/WarpSnapshotTests/Components/Native/TabBar/…`) and the house convention is 6-brand snapshot tests. A snapshot suite over the pre-26 path plus unit coverage of the identifier→index mapping would have caught the deselect issue I flagged inline, so it'd pay for itself here.

  ### Nits - all non-blocking, take or leave

  - `ScrollableTabview.swift` - filename typo, lowercase `v`, vs. the type `ScrollableTabView`.
  - `ScrollableTabview.swift:116-121` - `onSelect` fires from `shouldSelectItemAt`, which is a query method, as a side effect; `didSelectItemAt` is the more idiomatic home. Also the guard returns `true` when `indexPath.item >= items.count`, i.e. it permits exactly the out-of-range selection it looks like it's checking for.
  - A few raw literals where we have tokens: `ScrollableTabview.swift:97` `interGroupSpacing = 32` → `Warp.Spacing.spacing400`, and `:134` `indicatorHeight = 4` → `spacing50`. Also curious about `GlassSegmentedControl.swift:123` `pillBottomPadding = 3`, and `:44` `cornerRadius: 22` which `layoutSubviews` overwrites on the next pass anyway.
  - `GlassControlContainer.swift:3`/`:9` — internal `final class` with a `public init`. And the `registerForTraitChanges` handler at `:48-55` only re-applies `borderColor`, which is `.clear` with `borderWidth: 0` at the sole call site, so it's currently a no-op.

Sources/Components/Native/SegmentedControl/ScrollableTabview.swift
luk-m-s: Done

Sources/Components/Native/SegmentedControl/ScrollableTabview.swift
luk-m-s: Yup, move font solving inside the method, now it should be safe (however I think this scenario is very unlikely due to setting up global theme on the app starting)

Sources/Components/Native/SegmentedControl/GlassSegmentedControl.swift
luk-m-s: Nope, SwiftUI entry point is an addon. I have a dream that this entire component could performantly rewritten in SwiftUI but it looks like the answer is NO for the moment :/

So I decide to leave GlassSegmentedControl for UIKit consume and add SwiftUI container for it.  So if NMP uses updateShadow() for sth maybe it's better to leave it in place.

MarcinGie (approved):

Sources/Components/Native/SegmentedControl/SegmentedControl.swift
luk-m-s: I remove `fixedSize` from #Preview. Howeveer `uiView.systemLayoutSizeFitting(...)` also doesn't work here because it returns zero at this point 🤷‍♂️

Sources/Components/Native/SegmentedControl/GlassSegmentedControl.swift
luk-m-s: 👍

Sources/Components/Native/SegmentedControl/GlassSegmentedControl.swift
luk-m-s: Remove hardcoded padding. However this needs to adjusted manually in NMP app when we are merging releases

Sources/Components/Native/SegmentedControl/ScrollableTabview.swift
luk-m-s: Yup, fixed now 👍

luk-m-s:
### Theming doesn't go through `@Environment(\.warpTheme)`

Unfortunately yes. Only SwiftUI can go via environment injection - UIKit should go via global var. But this component is specific case, all other Warp components are pure SwiftUI with optional UIKit mapping - so this makes environment injection support natural. But here we have imported UIKit code with SwiftUI wrapper, completely other way around :/

I could modify UIKit part with provider injection but for now I don't think it gives you any benefit.

### Public API surface - a question, not a defect

Yes, UIKit facing API is must have here. NMP already use it directly, Warp add only SwiftUI syntactic sugar  wrapper on it. As I said above component internal logic is UIKit, it doesn't make sense to expose only SwiftUI when NMP already consumes it as UIKit

### Overlap with `Warp.Tabs`

I completely miss that Tabs is scrollable 👀 This is old component which have `webcentric` look but maybe I could use its SwiftUI logic to reimplemented GlassSegmentedControl completely in SwiftUI. Not sure about GlassContainer capsule thought 🤔 

### Tests

Yup, omission from my side. But TBH your examples are of little use of now because native Glass components do NOT render via GraphicsImageRenderer API (I've heard that this will change with iOS 27). I should have add them.

PR #229, opened 2026-08-21
FEP-50 iOS: Wrong font color for navigation bar prominent color in Liquid Glass

luk-m-s:
## Why
Liquid Glass design system handles button colors through tint colors, not explicit text color attributes. Hardcoded foregroundColor values conflict with the tint-based color application and prevent proper theming.

## What
Remove hardcoded foregroundColor attributes from navigation bar button styling:
- NavigationBar+Style.swift: Remove foregroundColor from plain and prominent button appearances
- UIBarButtonItem+Style.swift: Remove foregroundColor attributes from normal/highlighted states

### DIFF
--- a/Sources/Components/Native/NavigationBar/NavigationBar+Style.swift
+++ b/Sources/Components/Native/NavigationBar/NavigationBar+Style.swift
@@ -24,14 +24,12 @@ private extension UINavigationBarAppearance {
         
         let buttonAppearance = UIBarButtonItemAppearance(style: .plain)
         buttonAppearance.normal.titleTextAttributes = [
-            .foregroundColor: Warp.UIColor.token.text,
             .font: Warp.Typography.body.uiFont
         ]
         appearance.buttonAppearance = buttonAppearance
         
         let prominentButtonAppearance = UIBarButtonItemAppearance(style: .prominent)
         let prominentButtonAttributes: [NSAttributedString.Key: Any] = [
-            .foregroundColor: Warp.UIColor.token.textInverted,
             .font: Warp.Typography.title4.uiFont
         ]
         prominentButtonAppearance.normal.titleTextAttributes = prominentButtonAttributes
--- a/Sources/Components/Native/NavigationBar/UIBarButtonItem+Style.swift
+++ b/Sources/Components/Native/NavigationBar/UIBarButtonItem+Style.swift
@@ -32,22 +32,18 @@ extension UIBarButtonItem {
             self.style = .plain
             tintColor = tint
             setTitleTextAttributes([
-                .foregroundColor: Warp.UIColor.token.text,
                 .font: Warp.Typography.body.uiFont
             ], for: .normal)
             setTitleTextAttributes([
-                .foregroundColor: Warp.UIColor.token.text,
                 .font: Warp.Typography.body.uiFont
             ], for: .highlighted)
         case .primary:
             self.style = .prominent
             tintColor = primaryTint
             setTitleTextAttributes([
-                .foregroundColor: Warp.UIColor.token.text,
                 .font: Warp.Typography.title4.uiFont
             ], for: .normal)
             setTitleTextAttributes([
-                .foregroundColor: Warp.UIColor.token.text,
                 .font: Warp.Typography.title4.uiFont
             ], for: .highlighted)
         }


Sources/Components/Native/NavigationBar/NavigationBar+Style.swift
```diff
@@ -24,14 +24,12 @@ private extension UINavigationBarAppearance {
         
         let buttonAppearance = UIBarButtonItemAppearance(style: .plain)
         buttonAppearance.normal.titleTextAttributes = [
-            .foregroundColor: Warp.UIColor.token.text,
```
MarcinGie, line 27: With `foregroundColor` gone, this title colour now comes from the navigation bar's `tintColor` - and that's only set in the static entry point, at `:75`. The instance method (`:96-103`), `UINavigationItem.warpLiquidGlassStyle()` (`:140-145`) and `UIViewController.warpNavigationBarStyle()` (`:158-162`) all apply this same appearance without ever setting a tint, so a bar styled through any of those falls back to the system default and the title comes out blue instead of `token.text`.

  Before this change those paths were correct, because the colour travelled with the appearance object rather than the bar.

  The SampleApp hides it: `ContentView.swift:102` and `:154` call the global method at launch, so `UINavigationBar.appearance().tintColor` is already set and every demo screen inherits it. A consumer that only uses the per-VC API wouldn't be so lucky.

  I don't think there's an obvious one-liner here since `UINavigationItem` has no bar to reach for, so I'll leave the shape of the fix to you - but I'd rather not ship the regression on three public entry points.

Sources/Components/Native/NavigationBar/NavigationBar+Style.swift
```diff
@@ -24,14 +24,12 @@ private extension UINavigationBarAppearance {
...
             .font: Warp.Typography.body.uiFont
         ]
         appearance.buttonAppearance = buttonAppearance
         
         let prominentButtonAppearance = UIBarButtonItemAppearance(style: .prominent)
         let prominentButtonAttributes: [NSAttributedString.Key: Any] = [
-            .foregroundColor: Warp.UIColor.token.textInverted,
```
MarcinGie, line 34: Question rather than an objection: with the attribute removed, UIKit now picks the prominent label colour by contrast against the fill, so it's no longer any Warp token - not `textInverted`, not `textInvertedStatic`. That's exactly what your description proposes and it matches what `Button+Style.swift` already does on the SwiftUI side, so I'm inclined to agree.

  Has Yuri signed off on the design system handing that colour to UIKit? It's the kind of thing that's easy to agree on now and awkward to argue about later when a brand's primary background lands near the contrast threshold and iOS flips the label from white to black on us.

Sources/Components/Native/NavigationBar/UIBarButtonItem+Style.swift
```diff
@@ -32,22 +32,18 @@ extension UIBarButtonItem {
             self.style = .plain
             tintColor = tint
             setTitleTextAttributes([
-                .foregroundColor: Warp.UIColor.token.text,
```
MarcinGie, line 35: This file is the safe half of the change, in case it's useful to know: both branches set `tintColor` on the item itself (`:33` and `:42`), so the titles have something to fall back to.

  I checked `token.text` against `token.icon` across all six brands in `Sources/Colors/Tokens/*Tokens.swift` - identical in every one (`gray900` light / `white` dark) - so the `.default` case at `:35` and `:38` is a visual no-op. And the `.primary` case at `:44` and `:47`
  was the actual bug: it was setting `token.text`, i.e. near-black, over a `buttonPrimaryBackground` fill. Removing it is right.

  Only noting it because the diff makes all four look like the same change, and they aren't.

MarcinGie (approved):
Good catch on this one, and I agree with the diagnosis - Liquid Glass derives the prominent label colour from the fill, so a hardcoded `foregroundColor` on top of it was always going to fight the tint. It also brings the UIKit path in line with `Button+Style.swift`, which has been doing tint-only (`.glassProminent` + `.tint(colors.buttonPrimaryBackground)`) all along, so the two APIs now agree.

  I checked one thing you may have worried about: `token.text` and `token.icon` resolve to the same colour (`gray900` / `white`) in all six brands, so for the `.default` / plain button the fall-through to `tintColor` is visually a no-op. That half is safe.

  Two things I'd like sorted before merge, and one design question.

Sources/Components/Native/NavigationBar/NavigationBar+Style.swift
luk-m-s: This change is introduced to fix NMP regression (due to our handling of Liquid Glass theme). Dunno if this cause a regression in anything other in existence :)

Sources/Components/Native/NavigationBar/NavigationBar+Style.swift
luk-m-s: The source of the change was actual issue reported by Rita on Slack channel, Yuri was also a member. After long thread we all agree that sticking to the system behavior will be the best in long term.

Unfortunately it also means that all prominent/primary liquid glass buttons needs to be manually annotated in NMP codebase to get proper color. Search bar button fix is already commited.

Sources/Components/Native/NavigationBar/UIBarButtonItem+Style.swift
luk-m-s: 👍

PR #230, opened 2026-08-31
Raise minimum deployment target from iOS 17 to iOS 18

MarcinGie:
## Why

Warp's `Package.swift` declared `.iOS(.v17)`. NMP has been on an iOS 18 floor since 2026-06-16 (`ios-app` PR #10148, "Drop iOS 17 support"), so Warp was advertising support for a version no consumer actually targets.

Raised in the sheet/navbar thread by @Łukasz Śliwa: *"Warp package designation is misleading - it should be bumped anyway to `.iOS(18)`, NMP is done with it so we should do that too."*

Note this deliberately stops at 18 rather than 26. Bumping the whole package to `.iOS(26)` isn't possible while NMP is on 18. If we ever want native components to be strictly 26-only, the option discussed was a separate `warp-native` sub-target, which isn't worth doing right now.

## What

- `Package.swift`: `.iOS(.v17)` → `.iOS("18.0")`
- SampleApp deployment target: `17.6` → `18.0` (it consumes Warp, so it would otherwise fail to resolve)
- Removed three annotations the new floor makes redundant:
  - `Tabs.swift` `@available(iOS 17.0, *)` on a `#Preview`
  - `ColorFactorySmokeTests` `@available(iOS 17.0, *)`
  - `ThemePreviewModifier` `@available(iOS 18.0, *)` - this one was a real gate, keeping the theme preview traits (`.themeFinn`, `.themeBlocket`, …) unavailable to consumers below iOS 18. They're now unconditional.

### Why the string form

`SupportedPlatform.v18` requires `swift-tools-version: 6.0` and this package is on 5.9, so `.iOS(.v18)` doesn't compile. `.iOS("18.0")` is the equivalent that works on 5.9, and is what `ios-app` uses for the same reason.

## Verification

- `xcodebuild build -scheme warp-ios` on iOS 18.2 simulator: **BUILD SUCCEEDED**
- `xcodebuild test -scheme WarpTests` on iOS 18.2 simulator: **TEST SUCCEEDED**, 23 XCTest + 57 swift-testing tests in 8 suites, 0 failures

Snapshot tests were not run: they aren't part of CI and are unreliable locally, so they'd give a misleading signal either way. No snapshot references are touched by this change.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -6,7 +6,7 @@ import PackageDescription
 let package = Package(
     name: "warp-ios",
     platforms: [
-        .iOS(.v17)
+        .iOS("18.0")
     ],
     products: [
         .library(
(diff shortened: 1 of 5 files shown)


luk-m-s (approved):
🚀

PR #233, opened 2026-09-04
Remove dead pre-iOS-18 availability checks

MarcinGie:
## Why

#230 raised the package minimum to iOS 18. Every `#available(iOS 16.x)` and `#available(iOS 17)` branch left in the codebase is now statically true, and its `else` branch is unreachable. Leaving them in place makes the code read as though it still runs somewhere below 18.

## What

Removes all 12 of them: 8 in `Sources`, 2 in `Tests`, 2 in `SampleApp`.

Behaviour is unchanged. Two helpers collapse to nothing more than the call they were guarding, and the iOS 15 fallbacks they carried go with them:

- **`UnderlinedLinkModifier`** drew a 1pt line behind link text by hand, because `.underline(_:color:)` predates iOS 16. It was duplicated verbatim in `Box.swift` and `Alert.swift`; both are now a single `.underline` call.
- **`StepIndicatorView`** in the sample app carried two copies of the same `ScrollView`, the second one existing purely to omit `.scrollDisabled`.

Also drops `@available(iOS 16.0, *)` from `StateViewSheetModifier` and its `View` extension, and from two test methods.

## Tests

**TEST SUCCEEDED** on an iOS 18.5 simulator, 57 tests in 8 suites, 0 failures. The `Finn` SampleApp scheme builds.

### DIFF
--- a/SampleApp/Components/SelectView.swift
+++ b/SampleApp/Components/SelectView.swift
@@ -37,12 +37,8 @@ struct SelectView: View {
                     helpText: helpTextValue
                 )
                 .sheet(isPresented: $presentInfoPopover) {
-                    if #available(iOS 16.0, *) {
-                        createTooltipTextView
-                            .presentationDetents([.medium, .large])
-                    } else {
-                        createTooltipTextView
-                    }
+                    createTooltipTextView
+                        .presentationDetents([.medium, .large])
                 }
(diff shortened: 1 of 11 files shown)


luk-m-s (approved):
👍

PR #235, opened 2026-09-08
WARP-1451 Support iOS 18 to 25 in navigation bar style modifiers

MarcinGie:
## Why

Warp's navigation bar modifiers are iOS 26 only. Below that, `warpLiquidGlassStyle()` on `UINavigationBar`, `UINavigationController` and `UINavigationItem` returns early, `warpNavigationBarButton()` on `UIBarButtonItem` returns `self` untouched, and the SwiftUI `Button` modifier and `Warp.ToolbarItem` are gated out at compile time.

That was a deliberate decision, not an oversight: `8de39f4` and `7e9c2b7`, June 2026. It was the right call then. The package still declared `.iOS(.v17)`, and there was no agreement on what a pre-26 navigation bar should even look like.

Two things changed since:

1. The floor is now **iOS 18** (#230), and the iOS community meeting agreed Warp supports it.
2. Design answered the "how". @Adrian Dick and @Jonas Olsson landed on: no new Figma for pre-26, port what the apps already ship, QA by screenshot, useable rather than perfect.

Under that, an API that quietly does nothing across iOS 18 to 25 is no longer a scoping decision. It is a gap the consumer has no way to see.

@Łukasz Śliwa - this reverses your call, so it should be yours to sanity check. If you would rather own it or think the reasoning does not hold, say so and I will close this.

## Where the pre-26 appearance comes from

It is not invented here. NMP's `app-modules/Sources/UIAppearanceSetup/Appearance.swift` already has it, and its navigation bar setup is already shaped exactly like the thing we want to remove:

```swift
if #available(iOS 26, *) {
    UINavigationBar.warpLiquidGlassStyle()   // Warp
} else {
    ...hand-rolled appearance...             // outside Warp
}
```

Moving that `else` branch into Warp lets the call site collapse to one unconditional call. That is the clean move rather than an extra layer.

Its FinniversKit indirections all resolve to Warp already: `UIColor.background` and `.text` forward to `Warp.UIToken`, `UIFont.title1` and `.body` to the matching `Warp.Typography` cases, and `UIFont.bodyStrong` to `Warp.Typography.title4`. That last one is the same 16pt Medium `.callout` font as `Warp.Typography.bodyStrong`, but `title4` is what the chain resolves to and what the iOS 26 appearance already uses.

## What

- **`warpStyle()`** on `UINavigationBar` (static and instance), `UINavigationController` and `UINavigationItem`, each dispatching on the running OS. `warpLiquidGlassStyle()` is deprecated and forwards to it, so existing callers get a compiler warning rather than a silent behaviour change.
- **`UIBarButtonItem`** primary resolves to `.prominent` on iOS 26 and `.done` below, which is how NMP already expresses prominence there.
- **Ungates `Button.warpNavigationBarButton` and `Warp.ToolbarItem`**, which were iOS 26 only. Their sole 26 dependency was `.glassProminent`; `.borderedProminent` stands in below.

## Tests

`Tests/WarpTests/Native/NavigationBar/NavigationBarSolidStyleTests.swift`.

`UINavigationBarAppearance.warpNavigationBarSolid()` is `internal` rather than `private` on purpose. The public entry points dispatch on `#available`, so on an iOS 26 simulator the solid branch is unreachable and a suite going through them would pass without testing it. The new suite asserts on the appearance directly instead.

One existing test, `shouldDoNothingOnPreiOS26`, asserted the no-op on purpose. It now asserts the item is styled on every supported version.

**TEST SUCCEEDED** on iOS 18.5 and 26.5, 68 tests in 9 suites, 0 failures. The `Finn` SampleApp scheme builds and runs.

## Review note

Screenshots for design QA are worth attaching, but the honest caveat is that the visual delta on the sample app is subtle: the system default navigation bar was already white, so what actually changes below iOS 26 is the typography and the prominence of the primary bar button, not the background.


Sources/Components/Native/NavigationBar/NavigationBar+Style.swift
```diff
@@ -37,128 +37,198 @@ private extension UINavigationBarAppearance {
...
+    /// `UIColor.text` forward to `Warp.UIToken`, `UIFont.title1` to `Warp.Typography.title1`,
+    /// `UIFont.body` to `Warp.Typography.body`, and `UIFont.bodyStrong` to
+    /// **`Warp.Typography.title4`**, not to `Warp.Typography.bodyStrong`.
     ///
-    /// **Availability:** iOS 26+. On earlier versions, this method does nothing.
+    /// Internal rather than private so tests can assert on it from any OS. The public entry points
+    /// pick between this and the Liquid Glass appearance with `#available`, which would otherwise
+    /// make this branch unreachable from a suite running on iOS 26.
+    internal static func warpNavigationBarSolid() -> UINavigationBarAppearance {
+        let appearance = UINavigationBarAppearance()
```
luk-m-s, line 61: `internal` is redundant here

Sources/Components/Native/NavigationBar/NavigationBar+Style.swift
```diff
@@ -37,128 +37,198 @@ private extension UINavigationBarAppearance {
...
+        return appearance
+    }
+
+    /// The appearance for the running OS: Liquid Glass on iOS 26+, the solid appearance below.
+    static func warpNavigationBar() -> UINavigationBarAppearance {
+        if #available(iOS 26.0, *) {
+            return .warpNavigationBarLiquidGlass()
+        }
+        return .warpNavigationBarSolid()
+    }
```
luk-m-s, line 104: One caveat - not all navigation bars in NMP are translucent, if I remember correctly Drive proxy helpers already care about such case (you pass enum to helper with intention translucent/solid).

Here we don't assume such scenario which in theory can broke some screens which are not migrated to translucent navbar fully (dunno if this is a still a thing 🤷 )

luk-m-s (commented):
Just one comment about navbar styling compatibility with NMP integration

Sources/Components/Native/NavigationBar/NavigationBar+Style.swift
MarcinGie: Dropped in 4e54350.

It was there to signal "deliberately not `private`" - the public entry points dispatch on `#available`, so on an iOS 26 host the solid branch would be unreachable and the suite would pass without ever testing it. But the doc comment right above already says that in words, and the keyword the compiler ignores was the weaker half of saying it twice.

Sources/Components/Native/NavigationBar/NavigationBar+Style.swift
MarcinGie: Went and checked this against the NMP checkout, and I don't think this PR can break those screens - but the helper you're remembering is shaped a bit differently than described, so worth laying out the evidence.

**The helper.** It's `UINavigationController.applyStylingForModal(hideNavBar: Bool)` (`Drive/Public/Shared/Extensions/UINavigationController+Styling.swift`). It takes a `Bool`, not a translucent/solid enum, and the two appearances it picks between are `.opaque` and `.opaqueNoShadow` - both opaque, differing only in the shadow.

**Those statics derive from the global proxy rather than competing with it:**

```swift
static var opaque: UINavigationBarAppearance {
    let appearance = UINavigationBar.appearance().standardAppearance.copy()
    appearance.configureWithOpaqueBackground()
    appearance.backgroundColor = Warp.UIToken.background
    return appearance
}
```

(`Extensions/UIKit/UINavigationBarAppearance+Extensions.swift:15`) - it copies whatever `warpStyle()` installed, so modal bars inherit Warp styling instead of fighting it. I checked the one thing that looked risky there, whether `configureWithOpaqueBackground()` wipes the copied text attributes: it doesn't. Probed on a simulator against `warpNavigationBarSolid()`, and both `titleTextAttributes[.font]` (Inter-SemiBold 16) and `buttonAppearance.normal` survive the call untouched - it only rewrites background and shadow.

**Translucency specifically.** `isTranslucent = false` on the appearance proxy isn't something this PR introduces. NMP's own `Appearance.swift:29` sets it in the pre-26 branch today, and Warp's existing `warpLiquidGlassStyle()` on `main` sets it at line 78. So the translucency behaviour is unchanged for every caller, on both sides of the version split.

And where a screen does want something different, it sets it per-instance, which beats the appearance proxy. The one navigation bar in NMP that explicitly asks for translucency - `SchibstedTrust/.../DeleteFeedbackNavigationController.swift:20`, `navigationBar.isTranslucent = true` - is instance-level and unaffected by anything here.

If you had a specific screen in mind that I should look at directly, point me at it and I'll check that one rather than reasoning from the general case.

MarcinGie:
Status update, since CI here is red for reasons that are mostly not this PR.

**Merged `main` in** (`19d2a0d`) rather than rebasing, so the review threads above stay anchored. That picked up #238's toast fix, which is what the 09-18 red leg was.

**Three separate breaks were stacked on this branch's CI.** Two are now resolved:

1. **Toast wall-clock assertion** (`f43ac4c`). Fixed by #238, now in `main` via the merge above.
2. **ViewInspector 0.10.4**, published 2026-09-20. Its manifest declares `swift-tools-version:5.9` but uses `.visionOS(.v2)`, which needs PackageDescription 6.0, so it cannot compile for us and resolution dies with exit code 74 before anything builds. Nothing pinned it: the dependency was `from: "0.10.0"` and `Package.resolved` is untracked, so CI resolved fresh on every run and took 0.10.4 the day it appeared. This was breaking **every branch and `main`**, not just this one. Bounded in `660b38f` to `"0.10.0" ..< "0.10.4"`; lift it when upstream ships a release that compiles under 5.9.

**The third is pre-existing in `main` and is what still blocks this PR.**

The iOS 26 leg now gets past resolution and runs tests, then goes silent for 9 minutes inside `SheetStyleModifierTests` and is killed by the `timeout-minutes: 25` guard from #237. It reports as `cancelled`, which reads like someone pressed a button. It is not.

`main` does the same thing on its own head:

| head | run | last output | outcome |
|---|---|---|---|
| `main` @ `19e86c03` | [35388850899](https://github.com/warp-ds/warp-ios/actions/runs/35388850899) | `SheetStyleModifierTests`, 20:11:08 | cancelled at the 25 min timeout |
| this PR @ `660b38f` | [35695452489](https://github.com/warp-ds/warp-ios/actions/runs/35695452489) | `SheetStyleModifierTests`, 06:50:53 | cancelled at the 25 min timeout |

So a red iOS 26 leg on any PR right now is not evidence about that PR.

Worth flagging what those tests are, because the name misleads: they present no sheet and use no `ViewHosting`. They are pure value assertions that resolve colours heavily, through `UIColor(color).resolvedColor(with:)` in both interface styles across 6 brands and several styles. That is the same path as `ColorFactorySmokeTests testBackgroundColor`, the test that took 56.358s once and 0.232s the next run during the #237 triage. This looks like that same first-touch colour-resolution wedge, landing in a new suite because #220 added many more callers of it. `SheetStyleModifierTests` is a bare `@Suite` with no `.timeLimit`, so nothing bounds it and it eats the whole job.

**Where the navbar change itself stands.** It is complete. The CI-equivalent invocation locally, `xcodebuild -scheme WarpTests -destination "id=<udid>" -test-timeouts-enabled YES -default-test-execution-time-allowance 180 test`, passes **79 tests in 10 suites** on both iOS 26.0 and iOS 18.5. On CI, **the iOS 18 leg is green**. Only the wedged iOS 26 leg is outstanding.

Moving this back to draft while the wedge is chased in its own PR. Not asking for a re-review until it can go green.

MarcinGie:
Closing this. The change landed on `main` directly as `f8dba32` and `851f41b`, so this branch has nothing left that `main` does not already have - I diffed the trees to be sure. `main` is green on both legs.

@luk-m-s on your translucency point above: it went in without you having come back on it, which was not the plan for this one. The evidence in that thread still stands as far as I can tell - `applyStylingForModal` copies the proxy rather than competing with it, and `isTranslucent = false` was already set on both sides of the version split before this PR. But if you had a specific screen in mind, say which and I will check it against `main` and fix forward.

One correction to my status comment above, since it should not be left standing: the iOS 26 wedge described there is not a first-touch colour-resolution cost, and it was not specific to `SheetStyleModifierTests`. That diagnosis was wrong. The real cause is a synchronous, un-timed XPC call from `EnvironmentValues()` to the simulator's BoardServices daemon, which never returns on a bad boot. Fixed in #243, now on `main`.

PR #237, opened 2026-09-08
Test CI against the minimum supported iOS version

MarcinGie:
## Why

CI ran on `macos-26` against `platform=iOS Simulator,name=iPhone Air`. That resolves to the only runtime the image ships: **iOS 26**. The package declares **iOS 18** as its floor.

So every green check this repo has ever produced means "builds and passes on the newest OS". Nothing has ever verified the version we tell consumers we support.

That gap hides precisely the class of bug the floor exists to prevent: an API that compiles on every supported version and silently does nothing below the newest one. No warning, no crash, green tests. Both the recent navigation bar work (#235) and the sheet work (#220) turned up behaviour that only differs below iOS 26, and both were found by running two simulators by hand. The next person will not think to.

This is also what makes `PLATFORM_SUPPORT.md` (#236) enforceable rather than aspirational.

## What

Build and test as a matrix over iOS 18 and iOS 26, `fail-fast: false` so one version failing still reports the other.

### Why this moves off `macos-26`

The first attempt kept `macos-26` and downloaded the missing runtime. That does not work:

```
No iOS 18 runtime present, downloading it.
Finding content...
iOS 18 is not available for download.
```

`macos-26` ships iOS 26.2, 26.4 and 26.5 and nothing older, and an iOS 18 runtime cannot be added to it — `xcodebuild -downloadPlatform iOS` reports it as unavailable outright.

`macos-15` ships **iOS 18.5 and 18.6 alongside iOS 26.x**, so both legs run with no download at all.

Building still needs an iOS 26 SDK, because the source references iOS 26 symbols such as `.glassProminent` and `UIBarButtonItem.Style.prominent`. `macos-15` defaults to Xcode 16.4, which cannot compile those, so the workflow selects the highest installed 26.x toolchain — highest rather than a pinned point release, so an image update swapping 26.3 for 26.4 does not break the build.

Both legs share one toolchain, so a difference between them is attributable to the OS version rather than to the compiler.

**Trade-off:** this tests against Xcode 26.3 rather than the 26.6 available on `macos-26`. Verifying the floor is worth more than three point releases, and `macos-26` offers no way to have both.

### Other details

- **Resolves to a concrete UDID** rather than passing a device name. A bare `platform=iOS Simulator,name=...` destination can misresolve to a tvOS or visionOS device, and a name alone does not pin the OS version — which is how we ended up only testing 26 in the first place.
- **No download fallback.** A runtime missing from the image is generally also unavailable to download, so the job now fails with the available runtimes printed rather than attempting something that cannot succeed.
- **Boots the simulator before testing.** A first run against a cold simulator can report a phantom swift-testing failure that does not reproduce. Warming it avoids that without retry logic, which would mask real failures.

## Status

Green on both legs.

### The iOS 18 failure was real, and it is the first thing this PR caught

`testToastShouldAutomaticallyDisappear` and its snackbar twin set a 0.3s auto-dismiss and then slept 0.5s before asserting once, a 200ms margin. Dismissal is driven by a `.task` modifier, which only starts once SwiftUI has attached the hosted view, and on CI hardware that attachment alone can consume the margin. The tests were racing the runner rather than testing the component.

Replaced with a bounded poll: same assertion, returns as soon as the binding flips, and only spends the full timeout when something is genuinely broken. Faster too, since the common case no longer sleeps out the margin - the toast test goes from 3.83s to 0.71s.

Not an iOS 18 behaviour difference: the same test passes locally on an iOS 18.5 simulator. It is a latent race that only shows on slow hardware, and the matrix is what exposed it.

### The iOS 26 leg then wedged for six hours

On 2026-09-14 it hung inside `xcodebuild test` and ran until GitHub's 6-hour job ceiling killed it. That reports as `cancelled`, which reads like someone pressed a button. Nobody did.

The log stops immediately after the first test case starts:

```
09:41:24  ColorFactorySmokeTests testBackgroundColor started
15:37:55  ##[error]The operation was canceled.
```

That test is pure arithmetic over colour tokens - no async, no rendering, nothing but resolving a `Color`. It is simply first in the process, so it absorbs whatever one-time initialisation the first `Color.resolve(in:)` triggers. Everything after it runs in milliseconds.

The obvious theory was "iOS 26 runtime on the `macos-15` image", since the same test had taken **56.358s** on the iOS 26 leg three days earlier, against 0.129s on the iOS 18 leg and 0.019s on a local iOS 26.5 simulator. **That theory is wrong.** Re-run on the identical image build (`20260907.0337.1`) and the identical Xcode 26.3, the iOS 26 leg does that test in **0.232s**, indistinguishable from iOS 18. Nothing observable differs between the wedged run and the healthy ones, including the warm-up, which reports `Booted` in all of them.

So it is intermittent with no precursor visible from here, which is the case for bounding it rather than chasing it. Two bounds, doing different jobs:

- **`timeout-minutes: 25`** bounds the cost. A leg takes about 7 minutes, so this is an outer safety net rather than a budget.
- **`-test-timeouts-enabled YES -default-test-execution-time-allowance 180`** bounds the diagnosis. A test that overruns its allowance fails by name and with a spindump of the stuck thread, where a job timeout just stops the machine mid-sentence and leaves you reading the last line of the log.

Splitting the matrix across `macos-15` and `macos-26` would have been a plausible fix for the theory and no fix at all for the fact.

## Not in scope

`WarpSnapshotTests` still does not run in CI — 34 suites and 646 reference images, the only automated defence a six-brand visual system has. Tracked separately as FEP-137; a triage of why it is currently red is done and the references turn out to be stale rather than the code being broken.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -79,8 +79,19 @@ jobs:
         # Boot it up front. The first test run against a cold simulator can report a phantom
         # swift-testing failure that does not reproduce, so warm it rather than retrying failed
         # runs, which would mask real ones.
+        #
+        # Bounded poll rather than `simctl bootstatus -b`, which waits forever and hung a job here
+        # until the workflow timeout. Warming is an optimisation, so give up quietly and let
+        # xcodebuild boot the simulator itself if this does not finish.
         xcrun simctl boot "$UDID" || true
-        xcrun simctl bootstatus "$UDID" -b
+        STATE="unknown"
+        for _ in $(seq 1 30); do
+          STATE="$(xcrun simctl list devices -j \
+            | jq -r --arg u "$UDID" '[.devices[][] | select(.udid==$u)] | first | .state // "unknown"')"
+          [ "$STATE" = "Booted" ] && break
+          sleep 2
+        done
+        echo "Simulator state after warm-up: $STATE"
 
     # A UDID rather than a device name: a bare `platform=iOS Simulator,name=...` destination can
     # misresolve to a tvOS or visionOS device, and a name alone does not pin the OS version.
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -13,9 +13,65 @@ jobs:
 
     runs-on: macos-26
 
+    # Warp declares iOS 18 as its minimum (see PLATFORM_SUPPORT.md), so the floor is tested as
+    # well as the latest. Testing only the latest hides exactly the bugs the floor exists to
+    # prevent: an API can compile everywhere and silently do nothing below the newest OS.
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - ios-major: "18"
+            label: "minimum supported"
+          - ios-major: "26"
+            label: "latest"
+
+    name: iOS ${{ matrix.ios-major }} (${{ matrix.label }})
+
     steps:
     - uses: actions/checkout@v4
+
+    - name: Resolve an iOS ${{ matrix.ios-major }} simulator
+      id: sim
+      run: |
+        set -euo pipefail
+
+        find_udid() {
+          xcrun simctl list devices available -j \
+            | jq -r --arg v "${{ matrix.ios-major }}" \
+              '[.devices | to_entries[]
+                | select(.key | test("SimRuntime.iOS-" + $v + "-"))
+                | .value[] | select(.isAvailable)] | first | .udid // empty'
+        }
+
+        UDID="$(find_udid)"
+
+        # The runner image only ships the runtime bundled with its Xcode, so older ones have to
+        # be fetched. This is a multi-GB download, hence only when it is actually missing.
+        if [ -z "$UDID" ]; then
+          echo "No iOS ${{ matrix.ios-major }} runtime present, downloading it."
+          xcodebuild -downloadPlatform iOS -buildVersion ${{ matrix.ios-major }}
+          UDID="$(find_udid)"
+        fi
+
+        if [ -z "$UDID" ]; then
+          echo "::error::Could not find or install an iOS ${{ matrix.ios-major }} simulator runtime."
+          xcrun simctl list runtimes
+          exit 1
+        fi
+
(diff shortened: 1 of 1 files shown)


luk-m-s (approved):
Good backport - but just in future, bear in mind that Warp is hosted on free tier github account.

That just raises a question how much work we can push on CI side before we reach usage quota on workers 😅

MarcinGie:
Good flag, and the six-hour wedge I describe above is the concrete version of it, so it seemed worth pinning down the actual numbers rather than guessing.

**The minutes are fine, and this is the part that surprised me.** `warp-ds/warp-ios` is public, and GitHub Actions on *standard* GitHub-hosted runners is free for public repositories with no minute accounting at all. The 2,000 minutes/month on the Free plan is explicitly a **private**-repository quota, and the per-minute macOS rate only applies to billed usage. `macos-15` is a standard runner, so neither this matrix nor the six-hour wedge consumed any quota.

The line not to cross is **larger runners**, which are charged even on public repositories. So "how much work can we push" has an easy answer as long as we stay on standard runners.

**What is genuinely scarce is concurrency, not minutes.** The Free plan allows 20 concurrent jobs, of which **at most 5 may be macOS** (shared across standard and larger runners). `run-tests.yml` is macOS-only, so our real budget is five simultaneous jobs across the org. This matrix takes 2 of those 5 per PR for about 7 minutes, which means two PRs pushed at the same time already saturate it and a third queues.

That is the cost the wedge actually imposed: it held 1 of the 5 slots for six hours, and every macOS job in the org queued behind it for that long. Not a bill, a blockage. Which is what `timeout-minutes: 25` is for, and I would argue it is the single most valuable line in the diff for exactly the reason you raise.

Two follow-ons worth knowing:

- GitHub Support **can** raise the concurrency limit on request. The 6-hour job ceiling cannot be raised, which is why the wedge ran precisely that long.
- It is worth weighing against FEP-137. `WarpSnapshotTests` is 34 suites at roughly 5 to 6.5 minutes per run, so turning it on roughly doubles how long each PR occupies a scarce macOS slot. Still free, still worth doing, but the argument for it is about slot contention rather than money, and it is better made with that framing.

PR #238, opened 2026-09-18
Stop asserting dismissal timing that CI cannot measure

MarcinGie:
`main` is red on the iOS 26 leg. This fixes it.

## What broke

`f43ac4c` (merged in #237) added an elapsed-time assertion to the toast and snackbar dismissal tests. The reasoning was that the poll deadline alone could not tell a honoured 0.3s duration from a silent fallback to `Duration.short`, so the test would pass green while claiming to check something it no longer checked.

The threshold was 3s, calibrated against local runs of 0.5s warm and 1.1s cold. It broke `main` on the first post-merge run.

## Why retuning it does not work

Two consecutive CI runs:

| run | elapsed | dismissed |
| --- | --- | --- |
| `main` `cef9027` | 3.53s | true |
| #235 `4e54350` | 6.54s | false |

Hosting the view is what costs that, not the dismissal being measured. It is the same one-time initialisation that makes whichever test runs first in the process look pathological - the wedge investigation in `38bc852` saw `ColorFactorySmokeTests` take 56s on one run and 0.232s on the next, on an identical image and toolchain.

So the spread in attachment cost, 3.5s to 6.5s across consecutive runs, is the same order as the signal being detected: the gap between a 0.3s duration and the 5s fallback. Any threshold tight enough to catch a dropped `duration:` argument also fails on a slow runner, and any threshold loose enough to survive the runner cannot catch the argument. No value works, so the assertion is removed rather than retuned.

## Also raising the timeout

5.3s → 15s. On the second run the test was not merely failing the elapsed check, it was giving up before the toast dismissed at all. That would have been a false failure even without the extra assertion. The `.timeLimit(.minutes(1))` suite traits stay as the backstop.

## What this gives up

Real, and worth being explicit about: these tests now assert that the component dismisses, not that it dismissed *on the duration it was given*. A refactor that stopped forwarding `duration:` and fell back to `.short` would still pass.

Catching that needs a structural assertion on what the modifier forwards rather than a stopwatch. Both test files now say so at the line where the deadline is set, so the gap is documented rather than left for the next reader to rediscover.

## Verification

68 tests pass on an iOS 26.5 simulator.


luk-m-s (commented):
🚀

PR #239, opened 2026-09-18
FEP-115 Fix GlassSegmentedControl scroll view background bleeding through

luk-m-s:
## Why

`GlassSegmentedControl` wraps its segments in a `UIScrollView` to support horizontal scrolling when there are more segments than fit on screen. The scroll view's `backgroundColor` was set to `.clear`, which caused the segments' content to bleed through the transparent background as the user scrolled — content outside the visible bounds remained visible through the clear layer.

## What

Set `sv.backgroundColor` to `Warp.UIToken.background` instead of `.clear`.

The scroll view now paints the correct token background behind the segments, so content outside the visible clip area is hidden during scroll — matching expected behaviour.

### DIFF
--- a/Sources/Components/Native/SegmentedControl/GlassSegmentedControl.swift
+++ b/Sources/Components/Native/SegmentedControl/GlassSegmentedControl.swift
@@ -55,7 +55,7 @@ extension Warp {
         sv.showsVerticalScrollIndicator = false
         sv.alwaysBounceHorizontal = true
         sv.alwaysBounceVertical = false
-        sv.backgroundColor = .clear
+        sv.backgroundColor = Warp.UIToken.background
         sv.contentInsetAdjustmentBehavior = .never
         return sv
     }()


Sources/Components/Native/SegmentedControl/GlassSegmentedControl.swift
```diff
@@ -55,7 +55,7 @@ extension Warp {
         sv.showsVerticalScrollIndicator = false
         sv.alwaysBounceHorizontal = true
         sv.alwaysBounceVertical = false
-        sv.backgroundColor = .clear
+        sv.backgroundColor = Warp.UIToken.background
```
MarcinGie, line 58: The bug is real and well spotted, but I do not think this fixes it, and it costs the glass effect. Both checked on device, iOS 26.5, iPhone 17, sample app, with the control placed over a gradient so the behaviour is actually visible.

**It does not fix the bleed.** With `Warp.UIToken.background` applied and the control scrolled, the leading segment's glass capsule still renders outside the container - screenshot below, "New" escaping past the left edge onto the background. Same defect as on `main`, just on the other side.

That follows from where the escaping pixels are. A `backgroundColor` fills its own view's bounds; the capsule is being composited *outside* those bounds, so there is nothing there for the background to cover. iOS 26 Liquid Glass is composited by the system outside the normal layer tree, which is also why `clipsToBounds` on both `ControlScrollView` and `GlassControlContainer` does not bound it. This needs to stop the glass escaping, not paint behind it.

**It does remove the glass over content.** Detail and a before/after inline.

One note on why this is easy to get wrong in both directions: over the sample app's white `Form` row, none of this is visible. The escaping capsule is white-on-white and the occluded glass had a white backdrop anyway. I spent a while convinced the bug did not reproduce for exactly that reason. Anything we conclude here needs a non-uniform background behind the control.

Unrelated to the code: the failing checks are the ViewInspector 0.10.4 resolution break (`'v2' is unavailable`, from its visionOS declaration), which hits every branch right now. The pin is in #243 and #235; rebase once either lands.

| Image | Image |

luk-m-s: Ehh, yeah, you're right - it's glass on glass so I'll need some another approach to that :/

